PR #49, opened 2015-05-29
Closes drawer after menu selected if narrow

chuckh:
Adds onMenuSelect so the left sidebar menu closes after selecting a
menu item when drawerPanel is narrow.


addyosmani:
Thanks @chuckh!

PR #50, opened 2015-05-29
added simplified Chinese README link

unbug:
added simplified Chinese README link


unbug:
Don't understand why it's failed the test

addyosmani:
Could you squash your commits?

arthurvr:
Should we really add translations? Not IMHO. Mostly results in a pain later on. _Are they still up to date? Who will keep them up to date? Is the translation even good and correct?_

unbug:
@addyosmani Do you mean I should not do any PR like this?Of course.I understand,sorry about this ;)

PR #61, opened 2015-05-30
Avoid installing global dependencies and use npm scripts for quick start

mohsen1:
This will make it simpler to jump start with the project without having to install global dependencies. npm will use local `node_modules/.bin` folder for bower and gulp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,16 +25,17 @@ To take advantage of Polymer Starter Kit you need to:
 
 [Download](https://github.com/polymerelements/polymer-starter-kit/releases/latest) and extract Polymer Starter Kit to where you want to work.
 
-### Install dependencies
+### Quick start
 
-With [Node.js](http://nodejs.org) and npm installed, run:
+With [Node.js](http://nodejs.org) installed, run:
 
 ```sh
-$ npm run deps
+git clone https://github.com/PolymerElements/polymer-starter-kit.git
+cd polymer-starter-kit
+npm start
 ```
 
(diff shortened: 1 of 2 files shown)


addyosmani:
After digging into support for `npm` scripts on Windows, the npm2 implementation for `run` seems fraught with paths issues which is unfortunate. npm3 seems to tackle this better, but for now, I've switched to using the verbose form of installation instructions now available in the README to make this a little easier for Win users. 

Thanks for the PR anyway!

mohsen1:
AFAIK npm scripts has no issue in Windows. I use it in [Swagger Editor](https://github.com/swagger-api/swagger-editor/blob/master/package.json#L65-L73) with no issues. /cc @npm @othiym23

addyosmani:
That is really weird. Fwiw, things appeared to be smoother with the very latest npm 2.11.0, but that may just be me. For now the users who have been running into issues here appear to find using the commands directly to be less fragile.

mohsen1:
To be honest I had some bugs(links below) in Windows for npm scripts but recently nobody is reporting anything. 

https://github.com/swagger-api/swagger-editor/issues/424
https://github.com/swagger-api/swagger-editor/issues/407
https://github.com/swagger-api/swagger-editor/issues/390
https://github.com/swagger-api/swagger-editor/issues/270
https://github.com/swagger-api/swagger-editor/issues/216
https://github.com/swagger-api/swagger-editor/issues/132
https://github.com/swagger-api/swagger-editor/issues/425

PR #67, opened 2015-05-30
fixes #mainToolbar padding

kbwatts:
Otherwise {{appTitle}} is still visible when you scroll and flatten the waterfall-tall toolbar

### DIFF
--- a/app/elements/app-theme.html
+++ b/app/elements/app-theme.html
@@ -172,6 +172,13 @@ paper-menu a {
     padding-right: 30px;
   }
 
+    #mainToolbar.has-shadow .bottom {
+        margin-left: 50px;
+        padding-bottom: 0;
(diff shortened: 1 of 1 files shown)


addyosmani:
Thanks for the PR. Could you provide a before/after screenshot of how this addresses the issue? Locally, I'm unable to repro the title bar issue you're describing.

kbwatts:
As long as you're larger than the mobile view, this seems to happen at least on Chrome, Firefox, and Safari (just add some extra content to the initial page so there's enough to scroll):

before-chrome
before-chrome-detail
before-chrome-detail-dev
before-safari

After updating padding:
after-chrome
after-safari

addyosmani:
Changes LGTM! Thanks for bringing in the other changes too. Go team!

PR #77, opened 2015-05-31
Fixes #75 - correct text placement on scroll

### LINKED ISSUE #75: #mainToolbar text placement prevents hamburger menu from working on scroll
Once you add enough content to scroll the default waterfall-tall toolbar, the hamburger menu button seems to stop working - see screenshot (you can click but no effect - also affects mobile-sized/ios simulator):

hamburger-no-effect-wide
hamburger-no-effect-narrow
hamburger-no-effect-ios

addyosmani:
reviewer: @kbwatts 

I believe this corrects the issue you were experiencing:

screen shot 2015-05-30 at 20 17 58

### DIFF
--- a/app/elements/app-theme.html
+++ b/app/elements/app-theme.html
@@ -172,6 +172,12 @@ paper-menu a {
     padding-right: 30px;
   }
 
+  #mainToolbar.has-shadow .bottom {
+    padding-left: 50px;
+    padding-bottom: 0px;
+    font-size: 24px;
+  }
+
 }
 
 /* Material Design Adaptive Breakpoints */


kbwatts:
Looks good on Chrome, Firefox, and Safari :+1: 

However if you use margin-left instead of padding-left you'll also fix the hamburger menu bug (https://github.com/PolymerElements/polymer-starter-kit/issues/75) because it looks like the padding was overlapping on top of it. 

I updated my PR, but basically it's just updating your addition with:

```
#mainToolbar.has-shadow .bottom {
        margin-left: 50px;
        padding-bottom: 0;
        padding-left: 0;
        font-size: 24px;
    }
```

pdf:
@addyosmani as @kbwatts mentioned, `padding-left` does not actually fix the problem.  It must be `margin-left`, and in fact `margin-left` needs to replace `padding-left` for all breakpoints where `padding-left` is currently used.  I can submit a PR unless you're happy to fix this yourself.

addyosmani:
@pdf if there are remaining fixes required on top of https://github.com/PolymerElements/polymer-starter-kit/commit/3d7b50abc4a0281551d2dd0da1c3052cd3ef3a7f, a PR would definitely be welcome and appreciated.

PR #79, opened 2015-05-31
Fix hamburger menu button for additional media breakpoints

### LINKED ISSUE #75: #mainToolbar text placement prevents hamburger menu from working on scroll

pdf:
Fixes #75
Refs #77

### DIFF
--- a/app/elements/app-theme.html
+++ b/app/elements/app-theme.html
@@ -72,7 +72,7 @@ paper-menu iron-icon {
 
 #mainToolbar .bottom {
   padding-bottom: 120px;
-  padding-left: 30px;
+  margin-left: 30px;
 }
 
 paper-menu a {
@@ -150,7 +150,7 @@ paper-menu a {
   }
 
   #mainToolbar .bottom {
-    padding-left: 50px;
+    margin-left: 50px;
     padding-bottom: 24px;
   }
 


pdf:
So, my CLA is still broken since Google removed the ability to sign with an alternate email address (and removed my previously signed CLA along with it), as I've detailed on various issues on Polymer projects, without any result.

pdf:
As an aside, it'd be nice in the short term if there was at least an empty test so the CI wouldn't fail for PRs.

addyosmani:
Ugh. That sucks. I've just gone into our internal version of the system and I don't seem to have the ability to add a non-Gmail based address to it. Sorry about that @pdf. Is there another GMail/temp googlemail address you would be missing to use for the CLA? I hate to ask and will make sure to file feedback on the process needing to be more flexible.

pdf:
@addyosmani it used to be possible to add alternate addresses to your account and sign with that (which I did), in fact the [CLA form](https://cla.developers.google.com/clas/new?kind=KIND_INDIVIDUAL&domain=DOMAIN_GOOGLE) still links to [this article](https://support.google.com/accounts/answer/176347), but since the redesign of the Google account pages, those instructions don't work any more.

I guess could sign up for a Gmail account, then connect it to Github, sign the CLA with it, reconfigure and make sure I remember to configure each Google-related repository that I clone future to use that email address instead of the one I normally use for Github, then rewrite the history on any commits I've already made to them, but... it'd be nice if Google could un-break the CLA system instead.  I'm sure not everyone on Github uses a Gmail address for it.

addyosmani:
@pdf I'm told that the CLA has instructions for adding in alternative email addresses as long as the primary one is a gmail one. I completely appreciate that this might feel like unnecessary overhead. Right now there doesn't appear to me much we can do to change the process. Do you think you would be willing to complete the above? If so, that's great and we'd appreciate it. If not, I can rewrite the patch myself. Do you have a preference?

pdf:
Success!  I should apologise for not checking again - I'm not sure when it was released, but it looks like there's been another recent redesign of the Google Accounts UI and they've restored the ability to add an additional email address to an account, including non-Gmail accounts.  So, I've signed the CLA again.

addyosmani:
Yay! Thank you so much :star:

addyosmani:
Closed by 9e9516d7f22b853d1b3b7b16a6ae557df9f6ee5b

Issue #85, opened 2015-05-31
Duplicated lower bound for desktop x-large media query in the application theme


peterblazejewicz:
The media query breakpoints in application theme seems to be not very clear for x-large desktop breakpoint:
https://github.com/PolymerElements/polymer-starter-kit/commit/a8e5560c524a46201368d4efd944b2ecce807111#diff-a49e3b55ce6c03ad8616b1b64d5b5d1bR215
```
/* desktop-xlarge */
@media all and (min-width: 1601px) and (min-width: 1920px) { }
```
They are in line with material design but use doubled lower bound. I'm not sure if the intent is clear here.
Thanks!

addyosmani:
It should be min `1601` and max `1920` with an option to specify larger than that. I'll fix that up.

PR #86, opened 2015-05-31
Make desktop-x-large media queries more sensible

addyosmani:
Ref: https://github.com/PolymerElements/polymer-starter-kit/issues/85
Reviewer: @peterblazejewicz

### DIFF
--- a/app/elements/app-theme.html
+++ b/app/elements/app-theme.html
@@ -221,5 +221,5 @@ paper-menu a {
 /* desktop-large */
 @media all and (min-width: 1281px) and (max-width: 1600px) { }
 /* desktop-xlarge */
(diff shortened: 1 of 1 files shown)


peterblazejewicz:
@addyosmani LGTM
As you noted in the related issue the comment for optional upper bound could be helpful if added to source code, something like:
```html
/* optional desktop-xlarge breakpoint over 1920 */
```
and I would __not__ call it `xxlarge` as there is no such breakpoint in Material Design specs linked in the source comment.
Thanks!

addyosmani:
@peterblazejewicz I'm going to add in further information here based on the outcome of https://github.com/PolymerElements/polymer-starter-kit/issues/94. Input welcome there if you have any opinions :)

PR #88, opened 2015-05-31
Attempt to fix CI again

addyosmani:


addyosmani:
@cdata I'm probably being really dumb here. I'm attempting to get the Polymer CI server to attempt to test only the WCT config setup inside of /app/test, which locally gets run when trying `gulp test:local`. It appears CI wants those to exist in root/test or root/tests however. Any hints at how I can fix this up?

screen shot 2015-05-31 at 15 27 44

cdata:
That is a good question. Unfortunately, I don't have a good answer off of
the top of my head. I'll poke in WCT and see if there are any ways to
configure the runner statically.

cdata:
Okay, a quick peek into the WCT repo has yielded some actionable information.

First, the `Configuration` heading suggests that you can create a static configuration called `wct.conf.js`. Details here: https://github.com/Polymer/web-component-tester#configuration

Second, it looks like you can configure where to look for tests. Check this line in the sample configuration: https://github.com/Polymer/web-component-tester/blob/master/runner/config.js#L28

So, I've just taken a lot at `polymer-starter-kit`, and it looks like you already have a WCT configuration that has a value set for `suites`. Maybe I have misunderstood what the problem is - if so, could go into more detail about what you expect to happen and what is actually happening?

addyosmani:
Thanks for the reply! As you noted we currently have some WCT configuration set up in the repo to use the tests defined in app/test. Right now, the Polymer CI server fails citing `No test suites were found matching your configuration` and `WCT searched for .js and .html files matching: test/`, despite the setup otherwise working fine with WCT when we actually use the CLI. This doesn't cause any issues for end-users, but does mean that every pull request currently 'fails' the unit tests, which aren't actually being found or run by CI. 

I wonder if @azakus might know if there's further config we need to add here to get tests on PRs passing without the not-found exception.

addyosmani:
On looking through other repos in PolymerElements, I'm seeing the same issue of failing tests where we seem to enable the CI tests against PRs even when no valid tests are there or discoverable. @azakus sorry for the re-ping, but I'm curious if there's any way to get whitelisted for CI checks if we can't get it running against our tests in `app/test` (tried getting this working but failed)

azakus:
If I'm reading https://github.com/Polymer/web-component-tester/blob/master/runner/config.js#L28 correctly, it looks like you should set `suites: ["app/tests/"]` in your `wct.config.json`

azakus:
#138 looks like it works for polymer-ci :)

addyosmani:
Thanks for the PR, @azakus! Current pulls queue still shows tests failing (https://github.com/PolymerElements/polymer-starter-kit/pulls) but I'll assume that new PRs won't have this issue. Will file a new issue if they do. Appreciate the help!

azakus:
Yep, if those PRs were rebased with the config fixes, then they should test correctly.

PR #90, opened 2015-06-01
got rid of git://

karolklp:
npm install wasn't working when behind the http proxy


addyosmani:
I prefer this approach now that folks are actively using the project. Will land for the next release.

PR #91, opened 2015-06-01
A single line for installing deps causes problems on Ubuntu.

samcarecho:
On Ubuntu the following command must be executed with "sudo", otherwise the user gets an access rights error.

On the other hand, if the Ubuntu user issues the command using "sudo", Bower will complain.

My proposed solution is to divide the one liner command into node-deps and app-deps and move them back to package.json.

```
"scripts": {
    "node-deps": "npm install -g gulp && npm install -g bower && npm install",
    "app-deps": "bower install"
  }
```

I tested the modified installation instruction on Ubuntu 14.04 64 with NodeJS 0.12 from nodesource repository.

Everything seems to work fine, including serving the app, building the app and testing the app.

I haven't had the chance to test on different systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,18 @@ To take advantage of Polymer Starter Kit you need to:
 
 ### Install dependencies
 
-With [Node.js](http://nodejs.org) and npm installed, run the following one liner from the root of your Polymer Starter Kit download:
+With [Node.js](http://nodejs.org) and npm installed, run the following line from the root of your Polymer Starter Kit download to install the Node dependencies:
 
+Use "sudo" if needed (Ubuntu).
 ```sh
-$ npm install -g gulp && npm install -g bower && npm install && bower install
+$ npm run node-deps
+```
+
+Now install the app dependencies:
+
+Do not use "sudo" here.
+```sh
+$ npm run app-deps
 ```
 
 This will install the element sets (Paper, Iron, Platinum) and tools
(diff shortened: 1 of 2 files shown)


addyosmani:
Pardon my ignorance. Is the correct solution to work around sudo here not the same as other *nix systems and correct the permissions setup? https://docs.npmjs.com/getting-started/fixing-npm-permissions

samcarecho:
@addyosmani 

From my knowledge, it's not a good practice to mess with permissions. Also, according to the tests I executed this morning, all the moving and permission change hassle is not necessary, at least on Ubuntu.

There is a simple question that we can make to try to figure this out. Why would nodesource release packages specific for ubuntu that would require all that moving and permission setting hassle after installation? That does not make sense to me.

It could be that the npmjs doc is outdated or that it is just an hacky solution.

Let's put on hold  for now. I will test it again tonight on a fresh install of Ubuntu 14.04.
I will also try to test on other distros, if I manage to get the free time need.

Cheers

sindresorhus:
Addy is right. This is your solution: https://docs.npmjs.com/getting-started/fixing-npm-permissions

Alternative guide, with even an optional script that does it for you automagically:
https://github.com/sindresorhus/guides/blob/master/npm-global-without-sudo.md

Every project shouldn't have to work around your system. Doing the above will save you a lot of future trouble.

addyosmani:
I'm closing this with the above permission fix being the suggested solution. We've run into these challenges across Yeoman, Web Starter Kit and other npm-heavy projects and tend to find getting around needing sudo is the cleanest solution long-term.

PR #96, opened 2015-06-01
Update .bowerrc

dahuchao:
bower_components must be in the same directory that the index.html

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,3 +1,3 @@
 {
-  "directory": "bower_components"
+  "directory": "app/bower_components"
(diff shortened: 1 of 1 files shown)


addyosmani:
The current setup works fine for the non-light version of PSK. Can you confirm which version you're using based on the releases page? https://github.com/PolymerElements/polymer-starter-kit/releases

addyosmani:
We've decided to remove .bowerrc as part of the project and instead default to Bower's default installation targets. We'll document this once the change lands.

PR #98, opened 2015-06-01
min-aspect-ratio > landscape in app-theme

LM450N:


addyosmani:
Thanks for the PR. We hand wrote the original media queries, but obviously should have just straight-up used landscape for the above values. Landing.

LM450N:
I'm sorry, I should have to say "duplicated". It was late and like you can see, my English is a little bit poor. :wink: Thank you.

Issue #102, opened 2015-06-02
Version number in bower.json?


akhr:
polymer-starter-kit bower.json --> "version" : "0.0.0" should it be updated to "1.0.1" ?

addyosmani:
Good question. For our other boilerplate projects like Web Starter Kit, we opted not to bump the version number. This was intentional as it encourages end-users to think of the version field as theirs for their own projects. 

Interested in @wibblymat's take on this.

wibblymat:
Just to follow up: I was pretty sure that the `version` field in `bower.json` was never used, but that's not quite right. The only place that it is used is as the base version for the `bower version` command that bumps the version for you.

If you don't intend to use `bower version` for bumping the version it is safe to not include the `version` at all, since the tags are used instead.

PR #107, opened 2015-06-02
Mark packages as private

arthurvr:
Marks the NPM and Bower package private, and as a result removes all needless properties. This prevents people from accidentally publishing it to the registry.

Also, keeping those properties up to date is a waste of time when you're not planning to publish it and only are planning to use those for devDependencies (which out of the box is the case). There's no deal with publishing a whole site to bower/npm so I guess this is the right choice to make.

Fix https://github.com/PolymerElements/polymer-starter-kit/issues/102
Ref https://github.com/PolymerElements/polymer-starter-kit/commit/1076cd7fba3743aec12f98ddd3fec69a2aa691a5

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "polymer-starter-kit",
-  "version": "0.0.0",
-  "license": "http://polymer.github.io/LICENSE.txt",
+  "private": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


addyosmani:
Change LGTM. Thanks!

arthurvr:
My pleasure.

PR #120, opened 2015-06-04
Update `app.js` code style

arthurvr:
Generally I wouldn't create a PR like this but in boilerplates and starter kits this is
ends up being an inconsistency in not only this codebase but in the codebase of everybody
that uses that specific boilerplate.

Sorry :p


addyosmani:
:+1:

PR #127, opened 2015-06-05
Fix browser-sync in firefox after including paper-dialog (iron-overlay)

### LINKED ISSUE #126: Browser Sync Binding Breaks in Firefox
If paper-dialog is included, and thus iron-overlay-behaviour, Browser Sync tries to bind to the `<body>` tag in [iron-overlay's](https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L41]) documentation. This causes the rest of the file to get fubar'd and lots of weird application errors.

#127 "Fixes it on my machine"

zmoshansky:
Fixes #126.

I figured a span was a bit nicer than spamming the `@license` comment feature of vulcan. (I tried to use a comment to be less intrusive, but vulcan strips it out.)

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -56,6 +56,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 </head>
 
 <body unresolved class="fullbleed layout vertical">
+  <span id="browser-sync-binding"></span>
   <template is="dom-bind" id="app">
 
     <paper-drawer-panel id="paperDrawerPanel" forceNarrow>
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -182,6 +182,14 @@ gulp.task('clean', del.bind(null, ['.tmp', 'dist']));
 gulp.task('serve', ['styles', 'elements', 'images'], function () {
   browserSync({
     notify: false,
+    snippetOptions: {
+      rule: {
+        match: '<span id="browser-sync-binding"></span>',
+        fn: function (snippet) {
+          return snippet;
+        }
+      }
+    },
     // Run as an https by uncommenting 'https: true'
     // Note: this uses an unsigned certificate which on first access
     //       will present a certificate warning in the browser.
@@ -205,6 +213,14 @@ gulp.task('serve', ['styles', 'elements', 'images'], function () {
 gulp.task('serve:dist', ['default'], function () {
   browserSync({
     notify: false,
+    snippetOptions: {
+      rule: {
+        match: '<span id="browser-sync-binding"></span>',
+        fn: function (snippet) {
+          return snippet;
+        }
+      }
+    },
     // Run as an https by uncommenting 'https: true'
     // Note: this uses an unsigned certificate which on first access
     //       will present a certificate warning in the browser.


addyosmani:
Thanks for the PR! @shakyshane I'm unfortunately unfamiliar with how the snippet options config works here, but will test on our supported targets. Does this otherwise look above-board to you?

shakyShane:
The only issue I can see is that you may encounter unexpected results using that technique as the `<span>` element will remain in the DOM (the snippet will be appended, it will not replace by default).

Two better approaches would be

1. provide a custom replacement function to only return the snippet in *place* of the matched string (since `2.7.3`, you can provide a string for simpler replacements, no need for regex) - 

```js
require('browser-sync')({
    server: 'app',
    snippetOptions: {
        rule: {
            match: '<span id="browser-sync-binding"></span>'
            fn: function (snippet) {
                return snippet;
            }
        }
    }
});
```

2. use a html comment instead - this allows for a less configuration.

```
require('browser-sync')({
    server: 'app',
    snippetOptions: {
        rule: {
            match: "<!-- Browsersync -->"
        }
    }
});
```

zmoshansky:
Using a comment was my first thought as well, but it didn't work (as far my
testing shows) because Vulcan strips the comment out; IIRC, only @license
comments are left alone by Vulcan. I do like replacing the element as a
better fix.

addyosmani:
@zmoshansky IIRC, we don't run vulcanization during `gulp serve` so the comment being stripped out is less of an issue there - could you confirm/expand on your last comment? I may well be missing some context.

zmoshansky:
@addyosmani I didn't realize that `gulp serve` doesn't run vulcan; I've been using/must have tested `gulp serve:dist`. 

So moving forward, to fix the problem for `gulp serve`, a comment would work great; 
However, to also fix for `gulp serve:dist` a tag that gets replaced by browser sync would be the best solution, as suggested by @shakyShane.

I can update PR to either solution, depending on what the consensus is. My personal opinion would be to opt for a solution like @shakyShane's option 1 to ensure browserSync doesn't break when using `gulp serve:dist`.

addyosmani:
I think personally we should opt for option 1 from @shakyShane too :) Thanks for offering to update the PR!

zmoshansky:
@addyosmani Updated & Squashed.

addyosmani:
lgtm. @shakyShane unless there's further changes you would recommend to the above, I'll land this.

shakyShane:
Also you should probably bump Browsersync minimum version to `^2.7.7` in your deps as the string replacement feature was overhauled in [2.7.1](https://github.com/BrowserSync/browser-sync/blob/master/CHANGELOG.md#273-2015-05-24) so you'll want to make sure you're above that

addyosmani:
@zmoshansky want to bump browsersync as part of this or want me to bump after we land?

addyosmani:
Landed via b01c7986e76c60c16a1a67ded9604954472b3034. I'll bump BrowserSync separately. Thanks everyone <3

Bodhiz:
Hi guys,
Just experienced the issue, and try the fix provided here adding the ```<span>``` element and the SnippetOptions in my gulp task but BrowserSync still intercepts the ```<body>``` tag in the comment.  Am I missing something ? I'm running BrowserSync 2.8.2

PR #130, opened 2015-06-07
Layout and hamburger-menu improvements

addyosmani:
reviewer: @chuckh (if you have time)

This set of changes attempts to update the hamburger menu (and application name) positioning to be closer to the Material spec.

Spec:

Layout:

screen shot 2015-06-07 at 00 48 32

@kenchris is this closer to what you were after based on your reading of the spec? I think you were suggesting 20 was the right distance between the icon and title, however if it was 56px (your earlier number), that's easy enough to update to. 

screen shot 2015-06-07 at 00 57 56

It also includes several changes meant to improve the paper-header-panel/scroll behaviour of the layout, especially when there's more content than one card inside the page.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -117,7 +117,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
                 <p class="paper-font-body2">Looking for more Web App layouts? Check out our <a href="https://github.com/PolymerElements/app-layout-templates">layouts</a> collection. You can also <a href="http://polymerelements.github.io/app-layout-templates/">preview</a> them live.</p>
               </paper-material>
               <paper-material elevation="1">
-
                 <p class="paper-font-body2">This is another card.</p>
               </paper-material>
             </section>
@@ -147,8 +146,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             </section>
 
           </iron-pages>
-          </paper-header-panel>
-      </div>
+        </div>
+      </paper-header-panel>
     </paper-drawer-panel>
 
     <paper-toast id="caching-complete"
--- a/app/index.html
+++ b/app/index.html
@@ -85,6 +85,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           </paper-menu>
       </div>
       <paper-header-panel main mode="waterfall-tall">
+
         <!-- Main Toolbar -->
         <paper-toolbar id="mainToolbar">
           <button tabindex="1" id="paperToggle" paper-drawer-toggle>
@@ -103,6 +104,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         </paper-toolbar>
 
         <!-- Main Content -->
+        <div class="content">
           <iron-pages attr-for-selected="data-route" selected="{{route}}">
 
             <section data-route="home">
@@ -113,6 +115,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
                 <my-list></my-list>
 
                 <p class="paper-font-body2">Looking for more Web App layouts? Check out our <a href="https://github.com/PolymerElements/app-layout-templates">layouts</a> collection. You can also <a href="http://polymerelements.github.io/app-layout-templates/">preview</a> them live.</p>
+              </paper-material>
+              <paper-material elevation="1">
+
+                <p class="paper-font-body2">This is another card.</p>
               </paper-material>
             </section>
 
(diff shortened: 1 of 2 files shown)


addyosmani:
cc @arthurvr too :)

app/index.html
```diff
@@ -114,6 +116,10 @@
 
                 <p class="paper-font-body2">Looking for more Web App layouts? Check out our <a href="https://github.com/PolymerElements/app-layout-templates">layouts</a> collection. You can also <a href="http://polymerelements.github.io/app-layout-templates/">preview</a> them live.</p>
               </paper-material>
+              <paper-material elevation="1">
+
```
arthurvr: I have no idea what style guide we're supposed to use in here but this line looks ugly no?

app/index.html
```diff
@@ -142,6 +148,7 @@ <h2 class="paper-font-display2">Contact</h2>
 
           </iron-pages>
           </paper-header-panel>
+      </div>
```
arthurvr: The indentation here (this and the 2 lines above this) seems screwed up.

arthurvr:
Haven't dug into the MD spec at all, but the code looks good other than these 2 comments and everything tends to work fine.

app/index.html
addyosmani: This line is extraneous. Thanks for pointing it out. I'll nix it.

app/index.html
addyosmani: Confirmed. Will fix this! Thanks.

kenchris:
OK these are the values:

4 (padding left) + 12 (touch area left) + 24 (icon size) +  12 (touch right) + 4 (padding right) + 16 (padding to text)

or

16  + 24 (icon) + 16 + 16 then text. Distance from icon (without icon padding) is 32

Which is 72 - (left icon padding 16 + icon width 24) = 32

kenchris:
I don't get your layout image as the orange and yellow columns are overlapping the icon. They should not overlap the icon. The icon just doesn't contains iconography all the way to the edges of the image file which is fine :-)

kenchris:
Like this: https://drive.google.com/file/d/0ByMQtTUJSohfcjZuZEpzYVRmQ00/view

addyosmani:
As you've probably guessed, my design chops leave much to be desired :) Thanks for taking another look at the implementation. I'll update this per your suggestion shortly.

kenchris:
No problem :-) I worked closely with designers in a concepting team for years, so I can tell you that eventually it becomes second nature

addyosmani:
Updates:

screen shot 2015-06-08 at 13 10 26

screen shot 2015-06-08 at 13 09 54

@kenchris Having studied your graphic closer and spent time looking at paper-toolbar, I don't think that Starter Kit should be overriding the defaults it provides (beyond that for text) in order to match the spec. Rather, we should file bugs against paper-toolbar to be spec-compliant out of the box.

Some areas it could be improved:

* Spec suggests 56 as the height for the toolbar. paper-toolbar uses 64px. https://github.com/PolymerElements/paper-toolbar/blob/master/paper-toolbar.html#L84
* Spec suggests distance from icon to top/bottom of toolbar is ~16px. This isn't the case with paper-toolbar (and so on)

Do you think it's reasonable for us to land the improvements here as is and tackle the rest upstream?

chuckh:
@addyosmani @kenchris I do "think it's reasonable for us to land the improvements here as is and tackle the rest upstream such as toolbar". 

Reviewing the paper-toolbar demo at https://elements.polymer-project.org/elements/paper-toolbar?view=demo:demo/index.html only missing height was for Mobile landscape: 48dp. 64dp for desktop/tablet is correct as well as 56dp for mobile portrait.

## Metrics

**App Bar Default heights:**
Mobile Landscape: 48dp
Mobile Portrait: 56dp
Tablet/Desktop: 64dp

For extended app bars, the height is equal to the default height plus content increment(s).

From http://www.google.com/design/spec/layout/structure.html#structure-app-bar
App bar > Metrics

----
@addyosmani I'm getting this pull request, test and reviewing now.

kenchris:
I definitely think we should fix as much upstream in polymer related elements as possible. That is the right way to go.

But please make sure when everything is fixed upstream that it actually works in combination :)

kenchris:
@addyosmani Are you filing the bugs? Feel free to cc me

addyosmani:
@kenchris I'll either file the bugs or talk directly with the eng maintaining paper-toolbar about conducting an audit of spec-compliance. Will definitely cc you on what happens there in either case. I think you're definitely right that we should make sure these pieces work well when used together :)

@chuckh Thanks for reviewing! Do let me know if anything looks out of place :)

chuckh:
With the latest changes it looks good to me. :+1:  Nice work @addyosmani.

addyosmani:
Huge thanks to @chuckh for working through reviews via IM. Landing this set of changes :star:

PR #134, opened 2015-06-07
fixing some more indentation

arthurvr:
Fixing some more indentation ;)

On the while: it sucks that many existing tooling for javascript/css linting (and fixing) handles files, and not blocks inside of `<script>` and `<style>` tags.


addyosmani:
Thanks! Closed via 3f55ab5da2824046731c35dfc61e4817c50c36ba.

I also agree. There's a lack in tooling that correctly helps with many of the problems we face with inline content atm.

Issue #155, opened 2015-06-11
Moved <script> tags inside <dom-module> according to latest recommendations


chuckh:
Per latest Polymer release notes move `<script> tags` inside` <dom-module>`.

See release notes at https://www.polymer-project.org/1.0/docs/release-notes.html#documentation-updates-10-june-2015 

See @sjmiles example of this at http://jsbin.com/kobuvu/1/edit?html,output.

cnbuff410:
the "is: xxxx" property is not required officially?

addyosmani:
@chuckh good catch. We should be able to land a change for this to coincide with the 1.0.3 PSK release.

arthurvr:
Fixed in https://github.com/PolymerElements/polymer-starter-kit/commit/d7a146c8cb42cff3d7f4712eb381183fa827e79e?

chuckh:
Yes this now fixed with d7a146c and 5e682fa.

PR #161, opened 2015-06-11
move </dom-module> to bottom of file

yufengg:
I moved the </dom-module> closing tag to the bottom, and adjusted indentation. Github's diff shows it as being crazy.

Per the code samples in the docs (many examples, one such: https://www.polymer-project.org/1.0/docs/start/quick-tour.html), it seems like the style that is being encouraged is to have the dom-module wrap everything, including the <script> tag. If this is good, the my-greeting should also be updated similarly. It's worth noting that before the change to "fix indentation" on this file, all that needed to change was move the dom-module down, as the indentation was already set up for it to be at the bottom.

### DIFF
--- a/app/elements/my-greeting/my-greeting.html
+++ b/app/elements/my-greeting/my-greeting.html
@@ -26,19 +26,21 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <!-- Listens for "input" event and sets greeting to <input>.value -->
     <input class="paper-font-body2" value="{{greeting::input}}">
   </template>
-</dom-module>
-<script>
-  (function() {
-    Polymer({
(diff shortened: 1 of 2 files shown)


chuckh:
We also need to do this to https://github.com/yufengg/polymer-starter-kit/blob/patch-1/app/elements/my-greeting/my-greeting.html. Once this is done it will fix #155.

yufengg:
changes made to my-greeting.html

addyosmani:
lgtm

addyosmani:
Landed via https://github.com/PolymerElements/polymer-starter-kit/commit/d7a146c8cb42cff3d7f4712eb381183fa827e79e

PR #169, opened 2015-06-13
Delete `new-element.html`

arthurvr:
This file was empty and located on the wrong place. It didn't add any value.


addyosmani:
Thanks Arthur!

PR #180, opened 2015-06-17
Remove two moot `.gitignore` entries

arthurvr:
We don't have a `test/` directory, only `app/test` so these two don't really make sense.


addyosmani:
lgtm

PR #181, opened 2015-06-17
Remove confusing wording in `index.html`

arthurvr:
The usage of the wording "root directory" there might be confusing to people that are totally new to tooling. They actually need to add the icons to `app/` not really to the root `/` of their project. WDYT?


addyosmani:
lgtm. You're right that this updated phrasing is more clear.

PR #185, opened 2015-06-17
Fixes sub-title overlapping title (fixes #158)

### LINKED ISSUE #158: Application sub title does not work with waterfall-tall
In index.html there is a place to set an application sub title (not set by default) ([link](https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/index.html#L102-L103)):
```
<!-- Application sub title -->
<div class="bottom title"></div>
```

As you can see in the following animation application title and subtitle overlap when `paper-header-panel.mode="waterfall-tall"`takes effect (scrolling).
subtitle

Might want to not chose waterfall-tall, cut the subtitle from the template or refactor it to not be displayed if page-header-panel is collapsed.

chuckh:
Also changes all  uses of .has-shadow to :not(.tall). Fixes #158

### DIFF
--- a/app/styles/app-theme.html
+++ b/app/styles/app-theme.html
@@ -98,10 +98,20 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     margin-left: 48px;
   }
 
-  #mainToolbar.has-shadow .middle {
+  #mainToolbar:not(.tall) .middle {
     font-size: 20px;
     padding-bottom: 0;
     margin-left: 48px;
+  }
+  
+  #mainToolbar .bottom {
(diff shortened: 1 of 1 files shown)


chuckh:
Demo of these changes, though the pull request still leaves sub-title blank.

sub-title-fix.

We may in the future want to use the sub-title to show the current PSK version number. Maybe a custom-element could be created to pull the version number from the bower.json file.

addyosmani:
Thanks for the PR, @chuckh. Testing out now.

addyosmani:
Tested in our target browsers. Change lgtm for landing. Thanks!

PR #190, opened 2015-06-17
Remove redundant CSS height property. Closes #170

### LINKED ISSUE #170: a rule for #drawerToolbar in application theme overrides default rule for element height
PSK 1.0.2

I had a problem when applying `tall` class to `paper-toolbar` in the drawer when adapting default template to other project. Just adding documented classes seemed to not change element dimensions.

It appears that application theme declares property for element height that simply overrides defaults:
- PSK:
https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/styles/app-theme.html#L54
- `polymer-toolbar`:
https://github.com/PolymerElements/paper-toolbar/blob/master/paper-toolbar.html#L84

To use element classes as documented I had to remove that declaration from application theme rule. After this element uses the default height (the same 64px) but can be easily customized.

If that is OK, I could do PR,
Thanks!

peterblazejewicz:
This commit removes redundant height property declaration for
drawer paper toolbar component. Removing that property allows to
use documented classes to modify standard height.

After this commit the user would be able to use documented `tall` (or `medium-tall`) classes in application template to change `paper-toolbar` height.
Adding `tall` class:
```diff
          <!-- Drawer Toolbar -->
-        <paper-toolbar id="drawerToolbar">
+        <paper-toolbar id="drawerToolbar" class="tall">
           <span class="paper-font-title">Menu</span>
         </paper-toolbar>
```
- __before__ (no changes)

- __after__ (expected change):

Thanks!

### DIFF
--- a/app/styles/app-theme.html
+++ b/app/styles/app-theme.html
@@ -51,7 +51,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     color: var(--secondary-text-color);
     background-color: var(--drawer-menu-color);
     border-bottom: var(--drawer-toolbar-border-color);
-    height: 64px;
   }
 
   paper-material {


chuckh:
LGTM.

chuckh:
LGTM.

addyosmani:
LGTM

PR #151, opened 2015-06-10
Disable service workers by default

chuckh:
This is a first pass at a pull request  that does the following:
1. Disables services workers by default by using comments and removing precache
1. Updates readme.md to reflect these changes.

This addresses #80.

Issue #80, opened 2015-05-31
Make it easier for service workers to be optional


pdf:
For applications where it may not be possible to enforce the end-user (where the end-user is deploying the application) requirement of SSL deployment, or simply for those who don't want/need the extra complexity, it'd be nice if there was an easier way to disable service workers yet still sync the project (even if you're not syncing with this repo, removing chunks in half a dozen files is tedious).  This might possibly be handled by adding gulp tasks, and using gulp pre-processing to inject the SW stuff depending on which task is called, though the markup wouldn't be as clean.  Personally, I think I would prefer SW to be opt-in rather than opt-out, considering spotty browser support and added complexity.

arthurvr:
Well, I guess if we would ever base `generator-polymer` (the yeoman generator for polymer) on PSK, that would be something to consider. It would be as easy as a checkbox, then :)

addyosmani:
@pdf thanks for the feedback. The light version of Starter Kit (in the releaes page) doesn't include SW by default. 

I'd considered making it opt in in Starter Kit proper, but this would involve commenting out/in lines in about 4 different files (I've tried the code injection way before and it's far too fragile unfortunately) Side: I was hoping we could get more people doing offline-first by it being the default :)

Let us have a think more about how we can strike a balance between this being available without being too hard to either opt in or out. I'm not opposed to it being commented out.

@robdodson I agree that it might be useful to consider making SW support optional in the generator version of this project.

robdodson:
re: the generator. The plan at the moment is to base it on PSK by pulling in PSK through a git submodule and using that as the template. This is so PSK can be the canonical source. If it's easy to strip out the SW bits I'm happy to add it in, maybe they could have comments around them or something to make it easier to rip them out? I'm heading out for a week on vacation but happy to take a look when I get back

addyosmani:
I think for the V1 on the new generator, your plan sounds absolutely fine :) We may end up choosing to comment out SW support (would love it if it was still the default personally) but let's keep this issue open and see what folks think.

(also, thanks for updating the generator once again!)

calbertts:
:+1:

chuckh:
@robdodson regarding adding comments, would this make sense?

### index.html
```
    <!-- Service Workers -->
    <paper-toast id="caching-complete"
                 duration="6000"
                 text="Caching complete! This app will work offline.">
    </paper-toast>

    <platinum-sw-register auto-register
                          clients-claim
[...]

PR #191, opened 2015-06-17
Disable service workers

chuckh:
This is new pull request for disabling service worker. It disables it in code and updates readme.md for these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![](https://cloud.githubusercontent.com/assets/110953/7877439/6a69d03e-0590-11e5-9fac-c614246606de.png) 
+![](https://cloud.githubusercontent.com/assets/110953/7877439/6a69d03e-0590-11e5-9fac-c614246606de.png)
 ## Polymer Starter Kit
 
 > A starting point for building web applications with Polymer 1.0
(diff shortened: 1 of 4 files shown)


chuckh:
This pull request replaces pull request #151 and addresses #80.

addyosmani:
Thanks for the fresh PR, @chuckh!

In case anyone wonders why we've decided to switch Service Worker support to an opt-in model: at present, the way we lower the barrier to entry for setting up offline via `platinum-sw` is powerful, but also enables something of a catch-all for caching on the port you're serving from. 

What this means is that while you're fine using our SW strategy for PSK, as soon as you kill `gulp serve` and start using the port for something else, the same SW rules continue to apply, which is not ideal for end-users. You start getting unexpected exceptions in the console and really, if you haven't read at least a primer on using SW before you might not know how to reset/unregister to work around this.

Chrome are working on improving the DevTools story around SW and I'm hopeful we'll get to a point where it's very trivial to turn on and off SW support making the above issue a little more moot. Until then, PSK will stay offline-first with everything setup for you, but you'll need to uncomment SW-support in order to turn it on. 

We've weighed the pros and cons of outright removing the code, but have decided for the foreseeable future to keep it in there. SW support is huge for getting a clean offline story for webapps. We just want to make sure you're comfortable with enabling it should you need it.

addyosmani:
LGTM

PR #198, opened 2015-06-19
Use `npm test` script to run `test:local` task

arthurvr:
Having to remember the specific command in every different project is difficult so why don't we alias `gulp test:local` to the `npm test` script? Doesn't add too much value but can't hurt eighter. Just saves me personally some seconds every time I have to run the suite.


chuckh:
:+1:

addyosmani:
Change LGTM.

PR #203, opened 2015-06-22
Don't load files from bower_components

rubenstolk:


addyosmani:
For context, could you provide a description summarising the changes here to this PR please?

rubenstolk:
- Copies and loads webcomponents-lite.min.js to/from `elements/bootstrap` rather than depending on `bower_components`

IMHO we should remove `dist/bower_components` after vulcanizing but since people have reported that their apps use it, I've not included that in this PR (although I am doing that in my app for deployment).

filtercake:
(this PR btw enables deploying with Surge default settings – bower_components are ignored in default and webcomponentsjs then fails to load.)

rubenstolk:
Fixed 111 @filtercake thanks for the catch

addyosmani:
@rubenstolk can we squash the commits here?

rubenstolk:
Right, will do that, I'm not sure what went wrong when I was doing this all...

--
BR, Ruben Stolk
+919561050034  |  +31646118037




On Sat, Jun 27, 2015 at 1:01 PM -0700, "Addy Osmani" <notifications@github.com> wrote:












@rubenstolk can we squash the commits here?

gulpfile.js
```diff
@@ -96,8 +96,10 @@ gulp.task('copy', function () {
   var elements = gulp.src(['app/elements/**/*.html'])
     .pipe(gulp.dest('dist/elements'));
 
-  var swBootstrap = gulp.src(['bower_components/platinum-sw/bootstrap/*.js'])
-    .pipe(gulp.dest('dist/elements/bootstrap'));
+  var bootstrap = gulp.src([
+    'bower_components/webcomponentsjs/webcomponents-lite.min.js',
```
addybd: is it okay to hard code this path? if they ever change all project break and we might as well get it from bower.json `main`

rubenstolk: Isn't that the same issue with all other packages?

rubenstolk: There is another similar reference to webcomponents-lite.min.js in `gulpfile.js`

rubenstolk: squashed the commits by the way

rubenstolk: And by the way, I noticed one more file which is being tried by browsers `web-animations-next-lite.min.js.map`, trying to be loaded from `elements` directory.

In my local fork I've added this also to prevent 404's on my webserver:
```
var webanimations = gulp.src([
  'bower_components/web-animations-js/web-animations-next-lite.min.js.map'
]).pipe(gulp.dest('www/elements'));
```

(and adding `webanimations` to the `merge(...)` statement.

arthurvr: > if they ever change all project break and we might as well get it from bower.json main

When something like that happens the error is going to be quite clear so I don't see the need for overcomplicating the code therefore.

bppffnn: > When something like that happens the error is going to be quite clear so I don't see the need for overcomplicating the code therefore.

Please make it throw an error

addyosmani:
@arthurvr do you remember if we discussed this PR in our last team meeting? My notes don't show an entry for it :notebook:

arthurvr:
@addyosmani I don't think we did, no.

gulpfile.js
samccone: so the moment someone adds a new element they are going to have to add the new glob paths in here right?

hubgit:
When `dist/bower_components` is included, running `firebase deploy` on a vulcanized app takes a long time and uploads 1872 files; without `dist/bower_components` it takes a few seconds and uploads just 24 files.

Given that the app runs fine without `bower_components` folder (everything is imported by the `vulcanize` step) I'd like to be able to remove it from the distributed files.

addyosmani:
@samccone What are your thoughts on us including this patch? It's likely safe for the default setup in PSK and I agree that it does significantly impact deploy times in a positive way.

My only concerns are the edge cases where some dependencies may be dropped in and linked up by users but not properly copied over by Vulcanize (which I've run into before). We could suck it up and just say we'll file upstream bugs if we run into those cases however.

gulpfile.js
samccone: cc @addyosmani ^

samccone:
@addyosmani I had one outstanding question around new dependencies :)

gulpfile.js
addyosmani: @rubenstolk If you get a chance, we'd love to get your input on this question. This is the last outstanding question on the PR.

rubenstolk: Or we could simply go without naming them, in my current project this part looks like:

```js
return merge(

    gulp.src([
      src('/*'),
      '!'.concat(src('/test'))
    ], {
      dot: true
    }).pipe(gulp.dest(dest())),

    gulp.src([dir('/bower_components/**/*')])
      .pipe(gulp.dest(dest('/bower_components'))),

    gulp.src([src('/elements/**/*.html')])
      .pipe(gulp.dest(dest('/elements'))),

    gulp.src([src('/pages/**/*.html')])
      .pipe(gulp.dest(dest('/pages'))),

    gulp.src(['sc-elements/**/*.html'])
      .pipe(gulp.dest(dest('/sc-elements'))),

    gulp.src(['scripts/**/*.js'])
      .pipe(gulp.dest(dest('/scripts'))),

    gulp.src([src('/styles/**/*')])
      .pipe(gulp.dest(dest('/styles'))),

    gulp.src([src('/icons/**/*')])
      .pipe(gulp.dest(dest('/icons'))),

    gulp.src([src('/splash/**/*')])
      .pipe(gulp.dest(dest('/splash'))),

    gulp.src([
      dir('/bower_components/platinum-sw/bootstrap/*.js'),
      dir('/bower_components/webcomponentsjs/webcomponents-lite.min.js')
    ]).pipe(gulp.dest(dest('/elements/bootstrap'))),

    gulp.src([
      dir('/bower_components/web-animations-js/web-animations-next-lite.min.js.map')
    ]).pipe(gulp.dest(dest('/elements')))

  );
```

If that makes sense?

samccone:
Yeah I think the idea of having to modify your build file for every dependency change kinda stinks, in my mind, the gulp file should point at a folder and just pull everything in from it... sorta like it does now :)

Let's play this out a bit:

A user adds a new dep inside of their bower_components, they now have to go into the gulp file, change this, and then restart their watcher. 

If they forget to do any of the above steps things will not work and I think it will be confusing, and thus cause upstream issues to be filed. 

The point of a build file in my mind is something that you setup and do not have to think about in your normal dev flow... in a way this change totally breaks that mentality. 


-------

Onto your argument --- by using files right out of bower_components we are causing some people issues when they try and deploy, since many sites just ignore bower_components... You are right ... this totally stinks, however There is an entire suite of tools to bundle up your webapp and push it to N services ... aka this is a deployment problem not a dev (buildfile) problem

Sooo for now I am -1 to this PR landing, instead I would love to see some work around a better "ship" experience. 

Thanks a ton for your work on this @rubenstolk it is really appreciated

rubenstolk:
@samccone I don't think your completely right, with my PR it was never needed to explicitly embed any specific paths from bower_components. Vulcanize picks those up automatically and embeds them in your output and then bower_components is not needed. FYI I've been working on this branch for months and never saw the need for having bower_components in the build output. I don't think you should close it so lightly...

samccone:
hmm I can assure you I did not close it so lightly :) 
However I am totally open to further discussion

Could you expand on the wins we get with this "complexity"? 

Were this PR to land the advantage of being about to removing bower_components folder from our dist, does not outweigh the "cost" of this feature, which would be additional complexity around adding new bower_components (that are not <script> included inside of a vulcanized element)... but again totally open to being convinced otherwise, thanks for the response @rubenstolk :)

rubenstolk:
My only point would be that being a Polymer developer, after developing on this for a couple of months now, never saw the need of having bower_components in dist.

If something is not picked up by vulcanize, then it's probably an edge case and not bad to expect a developer to add that specifically, although even there I have never had to add any bower component to my gulpfile specifically.

Anyways, I don't fully understand the whole discussion I guess.

robdodson:
This PR hits on an interesting can of worms. I'll speak from my own experience if it's any help.

Generally speaking, vulcanize will most things and you can ignore the bower_components dir.

However, there are a handful of resources that need to be manually copied over from `bower_components`. `webcomponents-lite.js` is one. If you use `iron-ajax` and don't tell vulcanize to inline scripts, then the js associated with [promise-polyfill](https://github.com/PolymerElements/iron-ajax/blob/master/iron-request.html#L11) becomes another, if you're using platinum-sw register then the [bootstrap directory](https://github.com/PolymerElements/platinum-sw#relative-paths--vulcanization) becomes one (service worker scripts can't be inlined with vulcanize, they MUST be copied over), and if you use platinum-push-messaging then the [service-worker.js](https://github.com/PolymerElements/platinum-push-messaging/pull/60) becomes one.

So yeah, kinda annoying. The edge cases are pretty unavoidable.

The approach I've used on the last two sites that seemed to work ok was to only copy specific resources over using something like this in my `copy` task:

```
gulp.src([
  'app/static/bower_components/{webcomponentsjs,platinum-sw,sw-toolbox,promise-polyfill}/**/*'
])
  .pipe(gulp.dest('dist/static/bower_components'));
```

Otherwise I just let vulcanize do its thing and I don't copy over the `bower_components` dir. I think that's basically what this PR was aiming to do (although I don't agree with renaming the directory to `bootstrap` since that's specific to the service worker elements and they can be configured to look in a specific location or your server can rewrite the URLs which is what I do in `app.yaml`). I personally prefer this approach to just copying over all of bower_components. It isn't hard to maintain (I see a 404, I fix it) and maintaining this hasn't been complex or a burden on me.

samccone:
Hmm fair, in my mind the edge cases were hard to justify however I will defer to your well put thoughts and @rubenstolk's 

@rubenstolk mind reopening and we can land this? Thanks a lot for this work :)

robdodson:
Before landing I think it can be reworked some more. Unless I'm mistaken, [it's still copying the bower_components dir](https://github.com/PolymerElements/polymer-starter-kit/blob/867537b386e3663b97ceb62c1d867afd328f66c1/gulpfile.js#L94) (it's just not trying to load stuff from it)

Issue #214, opened 2015-06-24
Files in app/precache.json are discarded


robdodson:
I had imagined that if I added paths to `app/precache.json` that it would get merged with everything else in the [precache gulp step](https://github.com/PolymerElements/polymer-starter-kit/blob/master/gulpfile.js#L164) but that doesn't seem to be the case. Is this kind of thing discouraged or is it an oversight?

robdodson:
I wrote a PR @addyosmani @jeffposnick would either of you have a moment to review?

jeffposnick:
I want to make sure I understand the issue: are you manually adding resource paths to `precache.json` in the `app/` directory, and then you want to merge those manual additions with whatever gets picked up via globs in `dist/`?

If that's the use case, then have you considered just including the manually-managed paths as an array that you explicitly assign to the `precache` attribute of `<platinum-sw-cache>`? I'd consider that a cleaner approach, and that array is already automatically [merged](https://github.com/PolymerElements/platinum-sw/blob/master/platinum-sw-cache.html#L104) with the array that comes from `precache-file`.

Alternatively, if there are resource paths that you want precached that aren't being picked up by the existing glob patterns, maybe we can just change the default patterns or offer another way of customizing them?

robdodson:
> I want to make sure I understand the issue: are you manually adding resource paths to precache.json in the app/ directory, and then you want to merge those manual additions with whatever gets picked up via globs in dist/?

yes

> If that's the use case, then have you considered just including the manually-managed paths as an array that you explicitly assign to the precache attribute of <platinum-sw-cache>?

I did but there are a couple issues.

1. If you do put paths in `app/precache.json` they will be used when you run `gulp serve`. After you create a build using `gulp` they're no longer there. That part caught me off guard and seems like others would fall into it too.
2. I actually thought it was kind of nice being able to put the elements I wanted to cache in a separate json file. I preferred that approach to having it in my markup. I guess it's just personal opinion.

> Alternatively, if there are resource paths that you want precached that aren't being picked up by the existing glob patterns, maybe we can just change the default patterns or offer another way of customizing them?

[...]

PR #215, opened 2015-06-24
Fix Gh-214. Merge precaches

robdodson:
This allows the user to write files in their `app/precache.json` and have those merged with the file generated at `dist/precache.json`. This is useful for not having to go into the gulpfile to add files you want to precache.

Originally I ran into this issue when I added files to `precache.json` and ran `gulp serve` but was surprised when I generated a `dist` build and my files were no longer caching.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -21,6 +21,7 @@ var path = require('path');
 var fs = require('fs');
 var glob = require('glob');
 var historyApiFallback = require('connect-history-api-fallback');
+var arrayUniq = require('array-uniq');
 
 var AUTOPREFIXER_BROWSERS = [
   'ie >= 10',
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -21,6 +21,7 @@ var path = require('path');
 var fs = require('fs');
 var glob = require('glob');
 var historyApiFallback = require('connect-history-api-fallback');
+var _ = require('lodash');
```
arthurvr: Why are you pulling in the whole lodash to only use it once? Just use Sindre's [`array-uniq`](https://github.com/sindresorhus/array-uniq).

gulpfile.js
```diff
@@ -164,14 +165,27 @@ gulp.task('vulcanize', function () {
...
-    } else {
-      files.push('index.html', './', 'bower_components/webcomponentsjs/webcomponents-lite.min.js');
-      var filePath = path.join(dir, 'precache.json');
-      fs.writeFile(filePath, JSON.stringify(files), callback);
     }
+
+    var precache = JSON.parse(file);
+
+    // Build automated precache
+    glob('{elements,scripts,styles}/**/*.*', {cwd: dir}, function(error, files) {
```
arthurvr: Be consistent about the style of function definitions with the rest of the file. `function () {`.

gulpfile.js
```diff
@@ -164,14 +165,27 @@ gulp.task('vulcanize', function () {
 gulp.task('precache', function (callback) {
   var dir = 'dist';
 
-  glob('{elements,scripts,styles}/**/*.*', {cwd: dir}, function(error, files) {
+  // Read the development precache.json so we can merge it
+  fs.readFile('app/precache.json', 'utf8', function(error, file) {
```
arthurvr: Be consistent about the style of function definitions with the rest of the file. `function () {`.

gulpfile.js
robdodson: I was looking for a unique module but didn't find one. Thanks for that

gulpfile.js
robdodson: ok, the file actually already had function() {} case. I'll update

arthurvr: yeah we're inconsistent all over the place it seems :/

robdodson:
Updated

jeffposnick:
(See my comments in https://github.com/PolymerElements/polymer-starter-kit/issues/214)

gulpfile.js
addyosmani: Let's fix that up. We should probably be style linting against the gulpfile to avoid inconsistencies like that.

addyosmani:
Changes here lgtm, but would love to get a reply on Jeff's questions in #214 before we move ahead here.

addyosmani:
@arthurvr I'm going to do a code-style cleanup once this has landed.

addybd:
> Changes here lgtm, but would love to get a reply on Jeff's questions in #214 before we move ahead here.

ping

bppffnn:
LGTMMMM!!!!

bppffnn:
@addyosmani Crockford prefers 4 spaces over two maybe change that

addyosmani:
From Rob:

> After working on this for a while longer I'm fine closing the PR since I don't think it adds much value. If you were going to add only one file it's probably easier to do it with the attr and if you want to add multiple files it's probably better to use the globbing patterns in gulp.

We're working on an improved flow for working with precaching that we're hoping to have something ready to show from over the next few weeks. With that in mind we're going to close this up for now.

PR #216, opened 2015-06-25
Update to latest gulp

robdodson:
I started getting mismatch warnings today.

```
[19:46:59] Warning: gulp version mismatch:
[19:46:59] Global gulp is 3.9.0
[19:46:59] Local gulp is 3.8.11
```

If there's a reason to hold off on upgrading I'm cool with that. I didn't see a changelog in the gulp releases page so I'm not really sure the difference between 3.8 and 3.9 :\

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "connect-history-api-fallback": "^1.1.0",
     "del": "^1.1.1",
     "glob": "^5.0.6",
-    "gulp": "^3.8.5",
+    "gulp": "^3.9.0",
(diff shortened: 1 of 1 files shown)


arthurvr:
There's no need to update this, as we use [a caret](https://nodesource.com/blog/semver-tilde-and-caret). We're not using any 3.9-specific features.

addyosmani:
As @arthurvr mentions, the caret works sufficiently for us here. I would say we're also fine here given there's nothing 3.9 specific needed to run PSK.

PR #240, opened 2015-07-09
Fix paper-header-panel mode

tominon:


tominon:
I signed it!

flaiker:
I think this conflicts with #233.

addyosmani:
I've landed #233 as we're currently moving away from directly using paper-header-panel to using it's successor for the types of scroll interaction PSK's layout requires paper-scroll-header-panel. Afaik, this fix is no longer needed. Could you confirm?

(Also, thanks a lot for the PR. We appreciate the help!)

robbbz:
Need of a rebase.

chuckh:
Now that we are using `paper-scroll-header-panel` we can now close this. `paper-scroll-header-panel` does not support mode per docs at https://elements.polymer-project.org/elements/paper-scroll-header-panel?view=docs&active=paper-scroll-header-panel.

PR #247, opened 2015-07-11
enhance routing

naderio:
1- adds redirect to home when no path is matched (`page.redirect`)
2- removes redundant app.route value assignment

` app.route` is set twice when a menu item is clicked: 
* first by binding to `paper-menu`'s `selected` property
* second within the routing logic (where it should be)

3- resets main `paper-scroll-header-panel` scroll after route change (current behaviors is to leave scroll as is)

before changes:
before changes

after changes:
after changes

### DIFF
--- a/app/elements/routing.html
+++ b/app/elements/routing.html
@@ -40,6 +40,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       app.route = 'contact';
     });
 
+    page('*', function() {
+      app.$.toast.text = "Can't find: " + window.location.href  + ". Redirected you to Home Page";
+      app.$.toast.show();
+      page.redirect('/');
+    });
+
     // add #! before urls
     page({
       hashbang: true
--- a/app/index.html
+++ b/app/index.html
@@ -92,7 +92,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       </paper-scroll-header-panel>
 
       <!-- Main Area -->
-      <paper-scroll-header-panel main condenses keep-condensed-header>
+      <paper-scroll-header-panel main id="headerPanelMain" condenses keep-condensed-header>
 
         <!-- Main Toolbar -->
         <paper-toolbar id="mainToolbar" class="tall">
@@ -181,6 +181,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         </div>
       </paper-scroll-header-panel>
     </paper-drawer-panel>
+
+    <paper-toast id="toast">
+      <span class="toast-hide-button" role="button" tabindex="0" onclick="app.$.toast.hide()">Ok</span>
+    </paper-toast>
 
     <!-- Uncomment next block to enable Service Worker support (1/2) -->
     <!--
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -68,7 +68,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   // Scroll page to top and expand header
   app.scrollPageToTop = function() {
-    document.getElementById('mainContainer').scrollTop = 0;
+    app.$.headerPanelMain.scrollToTop(true);
   };
 
 })(document);
--- a/app/styles/app-theme.html
+++ b/app/styles/app-theme.html
@@ -127,6 +127,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     height: 900px;
   }
 
+  #toast .toast-hide-button {
+    color: #eeff41;
+    margin: 10px;
+  }
+
   /* Breakpoints */
 
   /* Small */


naderio:
I signed it!

app/elements/routing.html
```diff
@@ -12,6 +12,13 @@
 <script>
   window.addEventListener('WebComponentsReady', function() {
 
+    var headerPanel = document.querySelector('paper-scroll-header-panel[main]');
+
+    page('*', function(context, next) {
+      headerPanel.scroller.scrollTop = headerPanel.scroller.scrollTop <= headerPanel._headerMaxDelta ? 0 : headerPanel._headerMaxDelta + 1;
+      next();
+    });
+
```
chuckh: If @blasten thinks it safe to use `_headerMaxDelta` then I agree with this approach.

I think in the long run we will want each page to keep its' scroll position. If it is the first time on page then scroll to top with header expanded, then on following visits to page scroll to last position.

app/elements/routing.html
```diff
@@ -32,6 +39,10 @@
       app.route = 'contact';
     });
 
+    page('*', function () {
+      page.redirect('/');
+    });
+
```
chuckh: I like the idea of redirecting to home when no path is matched, but I think a paper-toast should popup with a message regarding the redirect which will make it friendly. Something like this:

image


```
page('*', function () {
      app.$.system_message.text = "Can't find: " + window.location.href  + ". Redirected you to Home Page";
      app.$.system_message.show();
      page.redirect('/');
});
```

And in index.html
```
<paper-toast id="system_message"
       duration="10000"
       text="Can't find url, redirected you to home page">
      <span role="button" tabindex="0" style="color: #eeff41;margin: 10px" onclick="app.$.system_message.hide()">Ok</span>
</paper-toast>
```

addyosmani: I like the idea of informing the user about the redirect via a toast. Wdyt @naderio?

blasten:
@addyosmani @chuckh I won't use `_headerMaxDelta`. From my understanding, it seems like you guys want a way to condense the header. is that correct? If that is the case, we can expose a method.

didrocks:
@blasten I +1 on the need of condensing the header. I had the exact same issue with Polymer 0.5 and did a quick hack in inheriting paper-scroll-header-panel at the time to have a forcecondensed property, that I used when navigating to a subpage (as I wanted the header to be condensed in non home page by default, but still being able to expand it)

chuckh:
>I won't use _headerMaxDelta. From my understanding, it seems like you guys want a way to condense the header. **is that correct?** If that is the case, we can expose a method in scroll-header-panel.

@blasten Yes.

addyosmani:
@blasten :+1: on a method. We could wrap up this PR if one was available and would appreciate it hugely.

blasten:
@addyosmani FYI: This methods are now available in paper-scroll-header-panel https://github.com/PolymerElements/paper-scroll-header-panel/releases/tag/v1.0.7

addyosmani:
Thanks for the heads up, @blasten! @naderio would you mind kindly updating your PR?

addyosmani:
Assigning myself as review. If for any reason @naderio doesn't get a chance to update the original PR, I'll rewrite this patch. Hoping for the former :stars: :space_invader:

naderio:
All, I apologize for my unresponsiveness!
I'll take care of this in a day or two at worst.
Thanks

arthurvr:
> All, I apologize for my unresponsiveness!
> I'll take care of this in a day or two at worst.

Thanks!

addyosmani:
Thanks!

arthurvr:
This will also need a rebase.

naderio:
updated .. rebasing

naderio:
Done.
@chuckh a slightly modified of your redirect toast notification have been added. Thanks!

app/elements/routing.html
```diff
@@ -12,6 +12,23 @@
...
+    var firstRouteChange = true;
+
+    page('*', function(context, next) {
+
+      headerPanel.scrollToTop(false);
+      if (headerPanel.headerState !== headerPanel.HEADER_STATE_CONDENSED && !firstRouteChange) {
+        headerPanel.condense(true);
+      }
+
+      if (firstRouteChange)
```
arthurvr: Missing braces

app/elements/routing.html
```diff
@@ -36,6 +53,13 @@
       app.scrollPageToTop();
     });
 
+    page('*', function () {
```
arthurvr: `function() {` for consistency with the rest of the file.

addyosmani:
@naderio thanks for working on the updates. Could you rebase against master and squash your commits please?

naderio:
@addyosmani rebased and squashed

app/elements/routing.html
```diff
@@ -12,15 +12,33 @@
...
+
+    var firstRouteChange = true;
+
+    page('*', function(context, next) {
+
+      if (!firstRouteChange) {
+        headerPanel.scrollToTop(false);
+        requestAnimationFrame(function(){
+          headerPanel.condense(false);
+        });
```
chuckh: `headerPanel.condense(false);` breaks the `app.scrollPageToTop` function. The header does not expand when changing pages anymore.  If you comment it out, it works.

app/elements/routing.html
```diff
@@ -31,11 +49,18 @@
...
       app.route = 'contact';
       app.scrollPageToTop();
     });
 
+    page('*', function() {
+      app.$.toast.text = 'URL not found! Redirected you to Home Page.';
+      app.$.toast.duration = 1000;
+      app.$.toast.show();
+      page.redirect('/');
+    });
```
chuckh: @naderio thank you work you work on this pr. You should lengthen the `duration to 4000` so there is enough time to read message.  You might also consider including the bad url in the message like below:
```
page('*', function() {
      app.$.toast.text = 'URL ' +  window.location.href  +  ' not found! Redirected you to Home Page.';
      app.$.toast.duration = 4000;
      app.$.toast.show();
      page.redirect('/');
    });
```

chuckh:
@naderio friendly ping regarding my line notes.

app/elements/routing.html
naderio: @chuckh changed

app/elements/routing.html
naderio: condensing the header after first view is the wanted behavior .. unless it isn't then removing `headerPanel.condense(false)` will do. even better we could use the smooth scroll with `headerPanel.scrollToTop(true)` in that case.

however, resetting the scroll *after* route change with repetitive `app.scrollPageToTop()` in every route function is redundant and probably not the best way to do it .. removed that.

`app.scrollPageToTop()` function is kept but its definition wa changed to a more suitable one.

chuckh: @naderio thanks for these updates! 

The purpose of `app.scrollPageToTop` function is to scroll to the top and expands the header. At the time of adding the function `document.getElementById('mainContainer').scrollTop = 0;` was the only way to do so. We decided to leave it for each route so users could choose which routes they wanted this behavior. We did not want to force every route to have this behavior. On some routes one might want to return to last position instead top for example.

naderio: ok, shall I revert lasrt changes?

app/scripts/app.js
```diff
@@ -68,7 +68,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   // Scroll page to top and expand header
   app.scrollPageToTop = function() {
-    document.getElementById('mainContainer').scrollTop = 0;
+    app.$.headerPanel.scrollToTop(true);
```
chuckh: In my testing `app.$.headerPanel.scrollToTop(true);` only scrolls to top but does not expand the header. So far I have only found `document.getElementById('mainContainer').scrollTop = 0;` scrolls to top and expands header. Which is desired behavior. So this should be reverted also.

app/index.html
```diff
@@ -88,7 +88,7 @@
       </paper-scroll-header-panel>
 
       <!-- Main Area -->
-      <paper-scroll-header-panel main condenses keep-condensed-header>
+      <paper-scroll-header-panel main id="headerPanel" condenses keep-condensed-header>
```
chuckh: I would recommend `id="headerPanel"` should be `id="headerPanelMain"`. Since we have two `paper-scroll-header-panels`. If we need an id for the other in the future it could be `id="headerPanelDrawer"`.  At this time I don't think we need either.

app/scripts/app.js
naderio: `bower install` please, should be fix by `paper-scroll-header-panel` update
`app.$.headerPanel.scrollToTop(true)` work as intended: scrolls to top smoothly then expands the header

chuckh: I tested again on another copy of PSK and `app.$.headerPanel.scrollToTop(true)` work!  @naderio thanks for sticking with it. :+1: 

So everything looks good to me.  If you want to rename headerPanel to headerPanelMain or not is ok by me.

naderio: @chuckh headerPanel renamed 
everything ok for me too, good to merge
thanks!

chuckh:
LGTM. Nice work @naderio!

chuckh:
@addyosmani everything looks good, do you want to land this PSK 1.1.0 or 1.1.2?

samccone:
hey @naderio sorry about the slow turn on this, would you mind rebasing once more

Thanks a ton!

naderio:
@samccone rebased

addyosmani:
Let's land this for 1.1.2. Thanks for rebasing!

Issue #237, opened 2015-07-06
routing.html page callbacks not called in iOS


kenjitayama:
How to reproduce

1. Download Starter Kit 1.0.2 or git clone the latest.

1. Edit routing.html and add these two cosole.log

		window.addEventListener('WebComponentsReady', function() {

		console.log('WebComponentsReady!');

		// We use Page.js for routing. This is a Micro
		// client-side router inspired by the Express router
		// More info: https://visionmedia.github.io/page.js/
		page('/', function () {
		  app.route = 'home';
		});

		page('/users', function () {
		  console.log('page users');
		  app.route = 'users';
		});

1. npm install && bower install

1. gulp serve

1. Browse to Users page from the drawer

The two logs are displayed in Mac Chrome/Safari but only the `WebComponentsReady!` is displayed in iOS 8.4 Safari (Confirmed by connecting from Mac Safari).
I asked my friend who has a Android to try it out, and looks like it's not called in Android too.

I'm creating a element for each page(contents of section tag), and I want to notify them when selected.
I'm doing something like the following, but it doesn't get called in iOS.

	document.querySelector('app-page-users').didShow(); // didShow is a manually implemented function

If this is a known issue due to browser compatibility, are there any work arounds for notifying my elements?

addyosmani:
I'll test this out on iOS and get back to you. Afaik, this should be working without issue.

kenjitayama:
Thanks!
I need this for loading data from the server at that timing.

kenjitayama:
I'd like to share a workaround that I came up with, although not ideal.

index.html
```
<section data-route="users">
  <app-page-users selected-page="[[route]]"></app-page-users>
</section>
```

AppPageBehavior.js (used in element app-page-users.html)
```
/* exported AppPageBehavior */
var AppPageBehavior = {
  properties: {
    selectedPage: {
      type: Boolean,
      observer: '_selectedPageChanged'
    }
  },

  _selectedPageChanged: function(newValue, oldValue) {

    var routeName = this.routeName;

    if (oldValue !== routeName && newValue === routeName) {
      if (typeof this._didAppear === 'function') {
        this._didAppear();
      }
    }

    if (oldValue === routeName && newValue !== routeName) {
      if (typeof this._didDisappear === 'function') {
        this._didDisappear();
      }
    }
  }
};
```

Not a good design because pages themselves should not be responsible for knowing which route is selected. But it works.
[...]

PR #256, opened 2015-07-17
fix routing.html page callbacks not called in iOS

kenjitayama:
fix for #237 

Needed to change the flow that comes before page.js callbacks

- before fix
`<a href="/pathForPage">` invokes page.js, but also invokes app.onMenuSelect() at the same time.

- after fix
`<a href="javascript:void(0)" path="/pathForPage">` invokes only app.onMenuSelect(event), and the corresponding page.js callback is called inside onMenuSelect.
Also changed `{{route}}` to `[[route]]` on the paper-menu (also included in PR #247)

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -69,18 +69,18 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         </paper-toolbar>
 
         <!-- Drawer Content -->
-        <paper-menu class="list" attr-for-selected="data-route" selected="{{route}}" on-iron-select="onMenuSelect">
-          <a data-route="home" href="/">
+        <paper-menu class="list" attr-for-selected="data-route" selected="[[route]]" on-iron-select="onMenuSelect">
+          <a data-route="home" href="javascript:void(0)" path="/">
             <iron-icon icon="home"></iron-icon>
             <span>Home</span>
           </a>
 
-          <a data-route="users" href="/users">
+          <a data-route="users" href="javascript:void(0)" path="/users">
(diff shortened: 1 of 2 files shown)


addyosmani:
cc @chuckh

chuckh:
@kenjitayama do you want to update your PR requests with the below changes?

remove  `on-iron-select="onMenuSelect"`
`<paper-menu class="list" attr-for-selected="data-route" selected="[[route]]">`

add ` on-click="onDataRouteClick"` to
`<a data-route="home" href="/" on-click="onDataRouteClick">`
`<a data-route="users" href="/users" on-click="onDataRouteClick">`
`<a data-route="contact" href="/contact" on-click="onDataRouteClick">`

Keep the original app.onMenuSelect code

```
app.onDataRouteClick = function() {
    var drawerPanel = document.querySelector('#paperDrawerPanel');
    if (drawerPanel && drawerPanel.narrow) {
      drawerPanel.closeDrawer();
    }
  };
```

This fixes the problem, plus hovering and right clicking to new tab works too.

_Note on-tap does not fix the problem._

cc: @addyosmani

kenjitayama:
@chuckh cc: @addyosmani 
Thanks! Updated PR accordingly.
I've also renamed onMenuSelect to onDataRouteClick.

chuckh:
LGTM!

addyosmani:
Thanks for the code review @chuckh and for working on this patch, @kenjitayama!

arthurvr:
@addyosmani If you land this way anyways, better to squash the commits :)

Issue #260, opened 2015-07-20
Broken 'paper-scroll-header-panel' condense effect when reusing the 'app-name', 'middle-container' or 'bottom-container' class


vguillou:
Just so you know, I ran into a small issue today by inadvertently reusing one of the CSS class of the main toolbar (namely 'bottom-container').
It broke the nice condensing effect of the paper-scroll-header-panel.

It's easily fixed by using a more specific selector in the main panel's scroll handler, like this

```
var appName = document.querySelector('#mainToolbar .app-name');
var middleContainer = document.querySelector('#mainToolbar .middle-container');
var bottomContainer = document.querySelector('#mainToolbar .bottom-container');
```

Pull request incoming

vguillou:
Should be closed now that f3bb4a3 landed. :+1:

arthurvr:
Yep, let's close it.

PR #261, opened 2015-07-20
#260 : Prevent the main panel's condensing effect from breaking when reusing the 'app-name', 'middle-container' or 'bottom-container' class

vguillou:
#260 : Prevent the main panel's condensing effect from breaking when reusing the 'app-name', 'middle-container' or 'bottom-container' class


chuckh:
LGTM

addyosmani:
LGTM2

PR #262, opened 2015-07-21
It adds gulp-minify-inline to the vulcanize

felixzapata:
It uses the gulp-minify-inline task to reduce the size of the vulcanized file.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -154,6 +154,7 @@ gulp.task('vulcanize', function () {
       inlineCss: true,
       inlineScripts: true
     }))
+    .pipe($.minifyInline())
     .pipe(gulp.dest(DEST_DIR))
     .pipe($.size({title: 'vulcanize'}));
 });
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "gulp-jshint": "^1.6.3",
     "gulp-load-plugins": "^0.10.0",
     "gulp-minify-html": "^1.0.2",
+    "gulp-minify-inline": "~0.1.1",
     "gulp-rename": "^1.2.0",
     "gulp-replace": "^0.5.3",
     "gulp-size": "^1.0.0",


felixzapata:
I signed it.

robbbz:
Vulcanize takes care of this.

felixzapata:
Hi, I don´t think so. (Or maybe I used the incorrect options). But if you try the default vulcanize tasks whith their options, the result file is 400kb. If you add the minifyOnline the result file is 200kb.

addyosmani:
So, the purpose of `gulp-inline-minify` is to uglify inline scripts. Instinctively, I would expect `inlineScripts` in Vulcanize to take care of this for me so it's possible I'm missing something. @azakus does this fulfil a need that Vulcanize doesn't currently already handle?

felixzapata:
I took the idea to do the pull-request after see how in [this article](https://aerotwist.com/blog/polymer-for-the-performance-obsessed/) the author use that task in the Vulcanize process. 

And after to review, the result of the vulcanize (with and without the minify task) in a couple of personal projects.

addyosmani:
Hah. Author of that article usually sits next to me. cc @paullewis - if the `inlineScripts` portion of the Vulcanize pipeline doesn't minify the inline scripts, we land land this.

felixzapata:
My opinion is that only with the Vulcanize task, I can't obtain the same results of size. As I say in a previous comment, maybe I am not using the correct options for Vulcanize, or maybe the Vulcanize has a bug.

The test is very easy. Run the vulcanize task with the minifyInline and without it and compare the size of the vulcanized file in both cases.

azakus:
No, vulcanize does not use uglify with `inlineScripts`. polybuild does with the `maximumCrush` option though.

felixzapata:
so, do I have to use ``.pipe(polybuild({maximumCrush: true})`` instead of my pull request code to obtain the same results?

azakus:
You'd get a bit closer. Polybuild doesn't include a CSS minifier, as I have yet to find one that handles the new custom property syntax (and polymer's mixin strategy).

addyosmani:
Thanks for the input @azakus. Now that Polybuild includes support for Gulp, should we be using it outright instead of Vulcanize?

azakus:
It would probably be worth experimenting with. Even the defaults of polybuild will remove whitespace indents in javascript, which is worth something.

felixzapata:
Hi, we have just discovered that with the minifyInline task (and using the theming syntax), the project does not load the css correctly. 

I think it is related with the previous comment of @azakus about the syntax.

So, sorry but I think my pull request is not 100% reliable (it depends of how to write your stylesheets and if you use theming syntax).

addyosmani:
After talking it over with @chuckh and @arthurvr we're going to experiment with polybuild in a branch before deciding how to move forward on this one. Goal is to have some learnings in the coming weeks.

addyosmani:
Closing in favour of https://github.com/PolymerElements/polymer-starter-kit/issues/312

addyosmani:
Thanks once again for your contribution @felixzapata. We appreciate it!

Issue #208, opened 2015-06-23
Add example of how to use babel or other transpilers


kenchris:
As many are using transpilers today, it would be nice with a some easy integration. Especially with a watch that transpiles on change, such that the web server is actually using the transpiled files.

Klathmon:
More than direct integration with one transpiler, i'd love a hook (or even a note in the documentation) where we can put a .pipe() and toss in whatever we want (babel, typescript, coffeescript, etc...)

cnbuff410:
Babel would be a decent starting point though.

addyosmani:
Imo, until Polymer has an official story around ES6, transpilation is something best covered in documentation or linking up to a repo demonstrating how to use Babel with Polymer Starter Kit.

arthurvr:
Ref https://github.com/yeoman/generator-polymer/issues/181.

> mostly team meetings/internal discussions. Let me try to pull up a summary of notes for you this week.

@addyosmani still interested :)

arthurvr:
> Imo, until Polymer has an official story around ES6, transpilation is something best covered in documentation or linking up to a repo demonstrating how to use Babel with Polymer Starter Kit.

Sounds good.

samccone:
https://github.com/PolymerElements/polymer-starter-kit/blob/master/docs/add-es2015-support-babel.md

documented! Thanks all for the interest!

PR #266, opened 2015-07-22
Added typescript 

gertcuykens:
You can use both type script and js script at the same time without extra build steps `gulp serve:dist`
Unlocks all features like in dept smart auto complete, type safety, definition look ups, and many many more of the visual code editor. ES6 apps right now that works in any browser.

```js
@component("my-greeting")
class MyGreeting extends polymer.Base {
   @property({notify: true}) 
   greeting:string = "Welcome!";
}
MyGreeting.register();
```
equals
```js
(function() {
  Polymer({
    is: 'my-greeting',
    properties: {
      greeting: {
        type: String,
        value: 'Welcome!',
        notify: true
      }
    }
  });
})();
```
Both will work, you can mix as much ts or non ts elements as you want.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ node_modules
 dist
 bower_components
 .tmp
+app/scripts/*.js
+!app/scripts/app.js
--- a/app/elements/my-greeting/my-greeting.html
+++ b/app/elements/my-greeting/my-greeting.html
@@ -26,21 +26,4 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <!-- Listens for "input" event and sets greeting to <input>.value -->
     <input class="paper-font-body2" value="{{greeting::input}}">
   </template>
-
-  <script>
-    (function() {
-      Polymer({
-        is: 'my-greeting',
-
-        properties: {
-          greeting: {
-            type: String,
-            value: 'Welcome!',
-            notify: true
-          }
-        }
-      });
-    })();
-  </script>
-
 </dom-module>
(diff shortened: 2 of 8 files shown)


arthurvr:
I'm :-1: on this. IMHO PSK shouldn't be opinionated what language (other than JavaScript) to use. Next time we get a request for Coffeescript, Opal, Livescript,... and all other languages. IMHO this is a better fit as a Yeoman generator that builds on top of generator-polymer.

gertcuykens:
Regular Javascript is valid TypeScript, it just extends javascript itself so we can use ES6 en beyond right now. I believe, like the angular2 developers it is going to take way to long for javascript it self to evolve where TypeScript is right now. I don't see it as just a other language but as javascript itself. Just look at it as a polyfill for javascript itself.

arthurvr:
@gertcuykens That's what everybody that uses some kind of preprocessor says too. We can't support all of them. Everybody likes their own thing and has their own opinions, that why I think this fits better as a yeoman generator that builds on top of generator-polymer. That would be fairly easy to use, though still optional.

I'd suggest you to share your thoughts on https://github.com/PolymerElements/polymer-starter-kit/issues/208, though.

> Just look at it as a polyfill for javascript ES6.

We [already](https://github.com/yeoman/generator-polymer/issues/181#issuecomment-108045948) decided that we didn't want that.

addyosmani:
Hey folks. We're trying to keep this project as lean (and pure) as possible as there are already quite a few moving pieces of machinery in PSK. A TypeScript variant makes most sense as a separate project or a tutorial that shows folks how to add TS support to PSK and provides a sample so that they can apply it to any version.

johnjelinek:
Has a TypeScript variant as a separate project or tutorial been created yet?

gertcuykens:
All polymer typescript work is done by https://github.com/nippur72/PolymerTS
The pull request is just a example how to use it.

PR #268, opened 2015-07-24
Change background of apple touch icon

### LINKED ISSUE #182: small issues with the Apple touch icon
Found two things regarding the touch icon:

* [We recommend to put `apple-touch-icon.png` in the root directory](https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/index.html#L20) while it actually lives in `images/touch`. The [`<link>`](https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/index.html#L40) is correct though.

* I always thought the logo was meant for a white background, while the icon is transparent and thus, looks like this on IOS:

screenshot_17_06_15_14_09

arthurvr:
Fix #182


addyosmani:
LGTM

PR #269, opened 2015-07-24
Gulpfile: add callback to `clean` task

arthurvr:
To let Gulp know when deletion of the folders is finished. This change recently happened in WSK as well: https://github.com/google/web-starter-kit/commit/6b4df236544cba9f115c731db721c8e865d5c6ad


addyosmani:
LGTM. Thanks for helping us stay in sync, @arthurvr!

PR #292, opened 2015-08-04
<platinum-sw> retooling

jeffposnick:
R: @addyosmani

This PR modifies PSK to take advantage of some of the changes that shipped with `<platinum-sw>` 1.2.0. My hope is it leads to a cleaner integration with PSK with fewer developer surprises/gotchas. I'm leaving the SW integration commented out for the time being, but I'd like to get to a point where we can ship PSK with it enabled by default.

Some notable changes:
- The default cache strategy is now @wibblymat's `sw-toolbox`'s [`fastest`](https://github.com/GoogleChrome/sw-toolbox#toolboxfastest) instead of [`networkFirst`](https://github.com/GoogleChrome/sw-toolbox#toolboxnetworkfirst). This will lead to better performance when making requests for cached resources, since they'll be served from the cache immediately, with the drawback of temporarily serving staler content even when online. Cached resources are effectively refreshed "in the background" whenever they're requested, ensuring that the next time they're requested, they're up to date. This is a common tradeoff when implementing caching.
- Instead of `precache.json`, containing just the array of files to be precached, the `gulp` task now outputs `cache-config.json`, which allows for customizing all of `<platinum-sw-cache>`'s options.
  - This makes it easy to disable caching completely in the `dev` environment, while enabling it in `dist`.
  - It also makes it possible to include a `"README":` in the static `app/cache-config.json` file, letting developers know not to modify the file. (This should help with the confusion that led to @robdodson's https://github.com/PolymerElements/polymer-starter-kit/pull/215)
- The generated cache config file relies on a `precacheFingerprint` to trigger the service worker upgrade flow when the list of files to precache changes, instead of serializing the file list as a service worker URL parameter. This works around [reported issues](https://github.com/PolymerElements/platinum-sw/issues/53) with a large number of precached files leading to invalid service worker URLs.
- It changes the default `dev` environment port to `5000` and `dist` environment port to `5001`.
  - This ensures that the `dev` and `dist` environments will no longer share a service worker.
  - It makes it less likely that other local unrelated development projects running in the same browser will accidentally pick up a PSK service worker, since `5000`/`5001` aren't common default HTTP ports.
- The project's name from `package.json` (falling back on the local directory name) will be used to create a distinct [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) object specific to that project, making it less likely for multiple PSK development projects opened in the same browser to accidentally share a cache. (This relies on developers updating their `package.json` with a meaningful value, though.)


addyosmani:
> I'm leaving the SW integration commented out for the time being, but I'd like to get to a point where we can ship PSK with it enabled by default.

Thanks @jeffposnick. This sounds reasonable and I think it's entirely desirable for us to evolve the SW implementation here to a point where it is low-friction enough to be prescribed by default.  

> The default cache strategy is now @wibblymat's sw-toolbox's fastest instead of networkFirst. This will lead to better performance when making requests for cached resources, since they'll be served from the cache immediately, with the drawback of temporarily serving staler content even when online.

I'm personally fine with us switching to the `fastest` approach. So, sgtm.

> Instead of precache.json, containing just the array of files to be precached, the gulp task now outputs cache-config.json, which allows for customizing all of <platinum-sw-cache>'s options.

As we've changing behaviour here, I think we should land this in a new minor version rather than just patch to indicate the shift in approach. Allowing us to fully disable caching in `dev` mode is a very welcome change.

> The generated cache config file relies on a precacheFingerprint to trigger the service worker upgrade flow when the list of files to precache changes

Could you walk me through a little more of how the fingerprinting setup works? It's good to be aware of https://github.com/PolymerElements/platinum-sw/issues/53 and that you're working around the issue.

> It changes the default dev environment port to 5000 and dist environment port to 5001

Fantastic! :star:

addyosmani:
Changes LGTM minus comments. cc @chuckh and @arthurvr as an FYI.

chuckh:
LGTM too.

jeffposnick:
>> The generated cache config file relies on a precacheFingerprint to trigger the service worker upgrade flow when the list of files to precache changes

> Could you walk me through a little more of how the fingerprinting setup works? It's good to be aware of PolymerElements/platinum-sw#53 and that you're working around the issue.

Sure.

Precaching during the very first SW installation on a given browser is straightforward, and the code uses the [`toolbox.precache()`](https://github.com/GoogleChrome/sw-toolbox#toolboxprecachearrayofurls) method to set the list of resources to precache prior to the install event being fired.

Whenever the list of files to precache changes, though (say, you've added a brand new subsection to your site), we want to be able to trigger the SW update flow, which will kick off a new `toolbox.precache()` -> SW install event to make sure the new content will be available offline immediately. There are two methods of triggering a SW install event: changing the content of the registered SW script file, or changing the URL of the registered SW script file.

Since `<platinum-sw-register>` uses a SW script file whose content never changes, we go with the approach of changing the URL of the registered script file to trigger the SW update flow. Specifically, we tweak the URL parameters.

Previous versions of `<platinum-sw-cache>` included a serialized array of all the files to be precached as part of the URL parameters. This had the intended effect of triggering the SW update flow whenever a new file was added, but had the unintended effect of creating untenably long URLs when there were a large-ish number of files to precache. To work around this, the code now includes a MD5 hash of the same serialized list of files as the URL parameter. This has the same beneficial property of triggering the update flow when the file list changes, while avoiding the URL length issue, since the MD5 hash uses a constant number of characters.

(I was going to initially call this parameter `precacheHash`, but [@wibblymat suggested](https://github.com/PolymerElements/platinum-sw/pull/57#issuecomment-126500283) `precacheFingerprint`, because there are other ways of achieving similar results beyond taking a hash of the serialized file list.)

addyosmani:
Lgtm as is. Thanks Jeff!

robdodson:
One quick question about the `fastest` strategy (sorry to ask after it's been merged). Will this cause someone to see stale files as they're developing?

jeffposnick:
Caching is disabled in the dev environment run via the `serve` task, i.e. `http://localhost:5000`. There shouldn't be an issue with stale files, and if there is, it's due to the browser's standard HTTP cache, not the service worker.

It's enabled in the dist environment run via the `serve:dist` task, i.e. `http://localhost:5001`.

robdodson:
Ok thanks Jeff

PR #303, opened 2015-08-10
removed broken reference for bower_components/page/page.js

simonfan:
The reference to the  bower_components/page/page.js was broken. It did not serve the component from the same route as the other bower_components.


addyosmani:
Are you using PSK light or the full version? There shouldn't be any issues with the page.js dep in either but will manually confirm once I know version.

addyosmani:
This should fix https://github.com/PolymerElements/polymer-starter-kit/issues/308 as well. LGTM.

simonfan:
thks

PR #304, opened 2015-08-11
✋ npm and global module gulp work 👷

samccone:
### What?

* Add npm cli is pretty awesome https://docs.npmjs.com/cli/run-script ... aliasing local package installs as npm commands is :cool: too.
* Since we only reference gulp via npm commands, people do not need to install gulp globally. :book: 

### Why?

Global modules can be problematic especially for users working on multiple projects with multiple versions of node and or gulp. This lets us not add to the community debt :bank: .

----

Let me know if this is no good, or a terrible idea, we can kill it :skull: no hard feelings :sparkling_heart: 

Thanks!!! :+1: :clap: :palm_tree:

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ This installs the element sets (Paper, Iron, Platinum) and tools the starter kit
 #### Serve / watch
 
 ```sh
-gulp serve
+npm run start
 ```
 
 This outputs an IP address you can use to locally test and another that can be used on devices connected to your network.
@@ -87,7 +87,7 @@ This outputs an IP address you can use to locally test and another that can be u
 #### Run tests
 
 ```sh
-gulp test:local
+npm test
 ```
 
 This runs the unit tests defined in the `app/test` directory through [web-component-tester](https://github.com/Polymer/web-component-tester).
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ The standard version of Polymer Starter Kit comes with tools that are very handy
 With Node.js installed, run the following one liner from the root of your Polymer Starter Kit download:
 
 ```sh
-npm install -g gulp bower && npm install && bower install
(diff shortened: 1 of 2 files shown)


addyosmani:
Lgtm. Will properly test when back from vaca but should be fine :)

arthurvr:
More of a personal thing, but I like using the Gulp CLI directly. Having the CLI installed globally has it's  advantages too (autocomplete in the shell, gulp-app, ...)

Also I think we should keep this inline with WSK, Yeoman gulp generators, ...

> We've tried to avoid folks using abstracted npm script commands in the past as we want them to know what gulp is doing

That's still a good point IMHO.

samccone:
> Global modules can be problematic especially for users working on multiple projects with multiple versions of node and or gulp. This lets us not add to the community debt

^ Is my primary argument. Assuming everyone can use a single version of a cli tool for all their local projects is not realistic in my findings in a real work multi project work environment

samccone:
> We've tried to avoid folks using abstracted npm script commands in the past as we want them to know what gulp is doing

I agree with the sentiment here, could the same be said about Makefiles ?

addyosmani:
After thinking about this a little further, I remember we had experimented with moving closer to `npm script` commands early on in our lifecycle. What we discovered was that while these commands work :dancer: on Mac and Linux, they caused a whole :truck: of issues for Windows users, even using the latest versions of npm and the early npm3. 

Our audience is also composed of quite a lot of beginners and intermediate developers who I feel would gain value from understanding the specific CLIs running under the hood (gulp etc). With that in mind, we're going to graciously turn down this PR, but think you're awesome for sending it over @samccone :tropical_drink:. We might be able to revisit this in the future if our userbase matures.

samccone:
No worries 👍👍

PR #309, opened 2015-08-15
Moving style tags to inside of template

alexpaluzzi:
New perf recommendation in v1.1

### DIFF
--- a/app/elements/my-greeting/my-greeting.html
+++ b/app/elements/my-greeting/my-greeting.html
@@ -9,18 +9,18 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 -->
 
 <dom-module id="my-greeting">
-  <style>
-    :host {
-      display: block;
-    }
-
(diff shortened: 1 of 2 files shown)


alexpaluzzi:
*Previously, we recommended that a `<style>` element should be placed inside an element’s `<dom-module>` but outside of its template. This is still supported, but we’ve now optimized placing styles within the template itself, so having the `<style>` tag outside of the template will be slower. We will be updating all of the elements provided by the Polymer team to match this new recommended behavior as well.*

https://blog.polymer-project.org/announcements/2015/08/13/1.1-release/

alexpaluzzi:
closing/reopening to re-run checks (non-code fail)

alexpaluzzi:
@addyosmani Not entirely sure why the check VM errored out, but it shouldn't be code related. Can we run again?

addyosmani:
I don't think the CI failure has anything to do with this PR. Merging away as is.

PR #311, opened 2015-08-17
Update dependency for Polymer 1.1.0

chuckh:
This is required for Polymer 1.1.0 to install. Also for Selenium to
work when running gulp test:local you have to update to Java 7 or
later. Updates readme.md for requirement.

@addyosmani please review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ gulp test:local
 
 This runs the unit tests defined in the `app/test` directory through [web-component-tester](https://github.com/Polymer/web-component-tester).
 
+To run tests Java 7 or higher is required.
+
 #### Build & Vulcanize
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ gulp test:local
 
 This runs the unit tests defined in the `app/test` directory through [web-component-tester](https://github.com/Polymer/web-component-tester).
 
-To run tests Java 7 or higher is required.
+To run tests Java 7 or higher is required. To update Java go to http://www.oracle.com/technetwork/java/javase/downloads/index.html and download ***JDK*** and install it.
 
 #### Build & Vulcanize
 
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "polymer-starter-kit",
   "private": true,
   "dependencies": {
-    "polymer": "Polymer/polymer#^1.1.0"
+    "polymer": "Polymer/polymer#^1.1.0",
     "iron-elements": "PolymerElements/iron-elements#^1.0.0",
     "paper-elements": "PolymerElements/paper-elements#^1.0.1",
     "platinum-elements": "PolymerElements/platinum-elements#^1.1.0",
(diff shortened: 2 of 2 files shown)


chuckh:
@addyosmani fixed missing comma, thanks. So far as of this morning nothing is requiring it yet. I checked iron-elements and paper-elements that don't (see https://github.com/PolymerElements/iron-elements/blob/master/bower.json for example).

I found with the styles moved inside template and you don't update Polymer 1.1.0 it breaks the `paper-scroll-header-panel` transforming of the `bottom-title`.

The test are breaking because it can't find Polymer 1.1.0.

notwaldorf:
@addyosmani @chuckh This is not needed. `^1.0.0` will pull anything up to `2.0.0`. The caret matches up to a major version change, the tilde up to a minor version.

chuckh:
@notwaldorf thanks for the explanation. Would you recommend changing the existing bower.json from various version numbers:

```
"iron-elements": "PolymerElements/iron-elements#^1.0.0",
"paper-elements": "PolymerElements/paper-elements#^1.0.1",
"platinum-elements": "PolymerElements/platinum-elements#^1.1.0",
"neon-elements": "PolymerElements/neon-elements#^1.0.0",
```

to all using only ^1.0.0

```
"polymer": "Polymer/polymer#^1.0.0",
"iron-elements": "PolymerElements/iron-elements#^1.0.0",
"paper-elements": "PolymerElements/paper-elements#^1.0.0",
"platinum-elements": "PolymerElements/platinum-elements#^1.0.0",
"neon-elements": "PolymerElements/neon-elements#^1.0.0",
```

notwaldorf:
@chuckh Yes. We (the elements team), prefer using 1.0.0 and trusting bower to Do The Right Thing™, but maybe they're like that for a reason. I defer to @addyosmani on that one :)

notwaldorf:
@chuckh gah update. I talked to @azakus and he says if we're relying on the new style features, we should use 1.1.0

chuckh:
@notwaldorf thanks for the followup. So we should use what I have in this pr then `"polymer": "Polymer/polymer#^1.1.0"` and leave the others as is?  

```
"polymer": "Polymer/polymer#^1.1.0",
"iron-elements": "PolymerElements/iron-elements#^1.0.0",
"paper-elements": "PolymerElements/paper-elements#^1.0.1",
"platinum-elements": "PolymerElements/platinum-elements#^1.1.0",
"neon-elements": "PolymerElements/neon-elements#^1.0.0",
```

addyosmani:
Thanks for the input, @notwaldorf and for checking in with @azakus. Unless I'm wrong, the proposal in @chuckh's last comment would be the right way to go here. 

Any disagreement @azakus?

README.md
```diff
@@ -92,6 +92,8 @@ gulp test:local
 
 This runs the unit tests defined in the `app/test` directory through [web-component-tester](https://github.com/Polymer/web-component-tester).
 
+To run tests Java 7 or higher is required.
```
arthurvr: Might be worth it to link people to further instructions on how to install (or update) Java.

chuckh: @arthurvr I added link with basic instructions to update Java.

PR #314, opened 2015-08-18
Adds support for polybuild (fixes #312)

### LINKED ISSUE #312: Explore support for polybuild
https://github.com/polymerlabs/polybuild

addyosmani:
r: @chuckh @azakus

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -147,18 +147,24 @@ gulp.task('html', function () {
     .pipe($.size({title: 'html'}));
 });
 
-// Polybuild will take take of inlining HTML imports,
-// scripts and CSS for you.
+// Polybuild will take care of inlining html imports,
+// scripts and css for you.
 gulp.task('vulcanize', function () {
-  var files = gulp.src('dist/index.html')
-  .pipe(polybuild({maximumCrush: true}))
-  .pipe(gulp.dest('dist/'));
+ return gulp.src('dist/index.html')
+ .pipe(polybuild({maximumCrush: true}))
+ .pipe(gulp.dest('dist/'));
+});
 
-  // Rename the generated index.build.html back to
-  // index.html as it's still your app entry point.
-  var index = gulp.src('dist/index.build.html')
-  .pipe($.rename('index.html'))
-  .pipe(gulp.dest('dist/'));
+// If you require more granular configuration of Vulcanize
+// than polybuild provides, follow instructions from readme at:
+// https://github.com/PolymerElements/polymer-starter-kit/#if-you-require-more-granular-configuration-of-vulcanize-than-polybuild-provides-you-an-option-by
+
+// Rename Polybuild's index.build.html to index.html
+gulp.task('rename-index', function () {
+ gulp.src('dist/index.build.html')
+ .pipe($.rename('index.html'))
+ .pipe(gulp.dest('dist/'));
+ return del(['dist/index.build.html']);
 });
 
 // Generate config data for the <sw-precache-cache> element.
@@ -256,12 +262,12 @@ gulp.task('serve:dist', ['default'], function () {
 
 // Build Production Files, the Default Task
 gulp.task('default', ['clean'], function (cb) {
-  // Uncomment 'cache-config' after 'vulcanize' if you are going to use service workers.
+  // Uncomment 'cache-config' after 'rename-index' if you are going to use service workers.
   runSequence(
     ['copy', 'styles'],
     'elements',
     ['jshint', 'images', 'fonts', 'html'],
-    'vulcanize', // 'cache-config',
+    'vulcanize','rename-index', // 'cache-config',
     cb);
 });
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -22,8 +22,8 @@ var fs = require('fs');
 var glob = require('glob');
 var historyApiFallback = require('connect-history-api-fallback');
 var packageJson = require('./package.json');
-var polybuild = require('polybuild');
 var crypto = require('crypto');
+var polybuild = require('polybuild');
 
 var AUTOPREFIXER_BROWSERS = [
   'ie >= 10',
@@ -147,8 +147,8 @@ gulp.task('html', function () {
     .pipe($.size({title: 'html'}));
 });
 
-// Polybuild will take take of inlining html imports,
-// scripts and css for you.
+// Polybuild will take take of inlining HTML imports,
+// scripts and CSS for you.
 gulp.task('vulcanize', function () {
   var files = gulp.src('dist/index.html')
   .pipe(polybuild({maximumCrush: true}))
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -146,19 +147,34 @@ gulp.task('html', function () {
     .pipe($.size({title: 'html'}));
 });
 
-// Vulcanize imports
+// Polybuild will take take of inlining html imports,
+// scripts and css for you.
```
arthurvr: `CSS` and `HTML`  for consistency.

gulpfile.js
```diff
@@ -146,19 +147,34 @@ gulp.task('html', function () {
...
+//   var DEST_DIR = 'dist/elements';
+//   return gulp.src('dist/elements/elements.vulcanized.html')
+//     .pipe($.vulcanize({
+//       stripComments: true,
+//       inlineCss: true,
+//       inlineScripts: true
+//     }))
+//     .pipe(gulp.dest(DEST_DIR))
+//     .pipe($.size({title: 'vulcanize'}));
+// });
```
arthurvr: Shouldn't this all be documentation (or in the Yeoman generator) instead of making the gulpfile even more noisy? Also big chunks of commented out code is a maintainability risk.

arthurvr:
Also fix the JSHint errors ;)

gulpfile.js
```diff
@@ -22,6 +22,7 @@ var fs = require('fs');
 var glob = require('glob');
 var historyApiFallback = require('connect-history-api-fallback');
 var packageJson = require('./package.json');
+var polybuild = require('polybuild');
```
arthurvr: How are these `require`'s ordered? The order seems strange to me. If there's no particular order let's add this one after `crypto`.

samccone: just wondering the thoughts as to why?

gulpfile.js
samccone: removing comments does provide a simple uncomment to go interface which can be quite nice as a consumer

arthurvr: An option in the Yeoman generator or a piece of documentation is way more nice to the consumer than a gulpfile cluttered with things only few consumers are going to use IMHO.

gulpfile.js
arthurvr: @samccone Because crypto is now the last one, and I was wondering why it got randomly inserted there instead of after the last one. Nitpicky though. In fact they just need to be ordered instead of having it this way ;)

gulpfile.js
arthurvr: And actually many consumers might not know what to comment out or change more when uncommenting this block.

gulpfile.js
samccone: if this is something we want to enforce someone should implement a linter to cause failures on CI otherwise this is a stylistic thing that becomes the burden of the reviewer to catch

addyosmani: There's really no reason. I'm happy to move it down :)

gulpfile.js
addyosmani: I'm fine with moving this code to the FAQ in case someone needs a copy/paste drop in that replaces the polybuild step.

samccone: :+1:

gulpfile.js
```diff
@@ -146,19 +147,34 @@ gulp.task('html', function () {
...
-      inlineScripts: true
-    }))
-    .pipe(gulp.dest(DEST_DIR))
-    .pipe($.size({title: 'vulcanize'}));
+  // Rename the generated index.build.html back to
+  // index.html as it's still your app entry point.
+  var index = gulp.src('dist/index.build.html')
+  .pipe($.rename('index.html'))
+  .pipe(gulp.dest('dist/'));
 });
```
chuckh: The renaming of `index.build.html` does not seem to work. To replicate run gulp, then check the `/dist/index.html`. Also notice you still have a `dist/index.build.html`.

I tried renaming `dis/index.html` to `/dist/index-org.html` as below. This did not work either. Should `dist/index.html` be delete first then rename?

```
  // Rename the generated index.build.html back to
  // index.html as it's still your app entry point.
  var indexOrg = gulp.src('dist/index.html')
  .pipe($.rename('index-org.html'))
  .pipe(gulp.dest('dist/'));
  var index = gulp.src('dist/index.build.html')
  .pipe($.rename('index.html'))
  .pipe(gulp.dest('dist/')); 
```

addyosmani: @chuckh Confirmed. If we throw a `del` in there to perform `del(['dist/index.html'])` before we attempt to rename, does that work for you?

chuckh: @addyosmani I tried `del`, no that does not work. The `del` and `rename` are being done async. So `del `and `rename` are happening before polybuild is done. So we may want to create a rename task that has a dependency on vulcanize task or maybe use run-sequence https://www.npmjs.com/package/run-sequence.

See also:
http://stackoverflow.com/questions/22824546/how-to-run-gulp-tasks-synchronously-one-after-the-other
http://blog.mdnbar.com/gulp-for-simple-build-proccess

addyosmani: Ah. They are. I'd vote for trying to use the rename task with a dep on the vulcanize task.

package.json
```diff
@@ -25,6 +25,8 @@
     "jshint-stylish": "^2.0.0",
     "merge-stream": "^0.1.7",
     "opn": "^1.0.0",
+    "polybuild": "^1.0.5",
+    "polyclean": "^1.2.0",
```
samccone: I think we should be able to drop this still :)

addyosmani: Now that the commented out block on Vulcanize is gone, we can nix polyclean.

samccone: woot!

gulpfile.js
chuckh: I got it to work by adding rename-index task, using return for vulcanize and rename-index tasks, and adding `rename-index` to the default task. @addyosmani you like to update this branch or would you like me to do so?

**vulcanize task changes, and new rename task:**
```
/ Polybuild will take take of inlining html imports,
// scripts and css for you.
gulp.task('vulcanize', function () {
  return gulp.src('dist/index.html')
  .pipe(polybuild({maximumCrush: true}))
  .pipe(gulp.dest('dist/'));
});

// Rename Polybuild's index.build.html to index.html
gulp.task('rename-index', function () {
  gulp.src('dist/index.build.html')
  .pipe($.rename('index.html'))
  .pipe(gulp.dest('dist/'));
  return del(['dist/index.build.html']);
});
```

**default task changes, added rename-index and updated comment to reflect change**
```
// Build Production Files, the Default Task
gulp.task('default', ['clean'], function (cb) {
  // Uncomment 'cache-config' after 'rename-index' if you are going to use service workers.
  runSequence(
    ['copy', 'styles'],
    'elements',
    ['jshint', 'images', 'fonts', 'html'],
    'vulcanize','rename-index', // 'cache-config',
    cb);
});
```

addyosmani: @chuckh Thanks for investing this further. That solution looks absolutely fine to me. I'm happy for either you to commit to the branch directly or for me to make these changes. Your call :)

chuckh:
@googlebot I authored https://github.com/PolymerElements/polymer-starter-kit/commit/b2240482061daefd9d2c85be76ab27f610672607 commit. I okay with this commit being contributed to Google. 

@addyosmani please review my latest commit.

marioAnfiled:
I did not author or contribute. In my effort to learn about github and Polymer I may have submitted something by accident.

addyosmani:
Tested locally and these changes LGTM. Let's land this :boat:

Issue #202, opened 2015-06-22
Content not scrolling back after navigating


Felizardo:
Hello,

I have noticed that the content div does not scroll back after navigating to other sections of the example. See image below.

![polymer_starter_kit_not_scrolling_back](https://cloud.githubusercontent.com/assets/1034631/8274621/82950f14-186b-11e5-89b7-65e6fdb3de3c.gif)

I could reproduce that on my Mac using Chrome, Safari and Firefox, and also on Chrome in my Android Lollipop device.

Felizardo:
FYI, a workaround is to comment the height attribute of .content on app-theme.html

addyosmani:
@Felizardo Thanks for the bug report. Could you verify if this is still an issue on Master?

Felizardo:
Hello @addyosmani, yes, I just made a clone right now and the problem persists.

DannieBGoode:
Work around until it gets fixed:

setting manually document.getElementById("mainContainer").scrollTop = 0 will scroll back up.

chuckh:
Pull Request #316 fixed this issue. See details of change at https://github.com/PolymerElements/polymer-starter-kit/pull/316/files.  It is in master branch now, it will be included in the next release 1.1.0 of Polymer Starter Kit.

Issue #285, opened 2015-07-29
Paper-scroll-header-panel


SayChunKim:
When switching the pages through menu, the content  doesn't reset back to initial height and paper-scroll-header-panel doesn't back to form that it hasn't condenses. Any solutions/workaround for now?

Version: 1.0.3

flaiker:
See #202, a workaround is mentioned

addyosmani:
From that issue: 

> setting manually document.getElementById("mainContainer").scrollTop = 0 will scroll back up.

Let's file this as a bug and aim to land a workaround/fix for 1.0.4.

SayChunKim:
SGTM. Thanks ! @addyosmani

arthurvr:
Can we close this now that https://github.com/PolymerElements/polymer-starter-kit/commit/ca0a9fca55d9add799e1d8b272b77acb0ff7f0b1 landed?

addyosmani:
I believe we can, yes. Thanks for pinging this issue, @arthurvr. Always helpful :sparkles

PR #316, opened 2015-08-20
Scroll to top when changing pages (fixes: #202, #285)

chuckh:


app/scripts/app.js
```diff
@@ -66,4 +66,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
   };
 
+  // Scroll page to top and expand header
+  app.scrollPageToTop = function() {
+    document.getElementById('mainContainer').scrollTop = 0;
```
samccone: any reason why not
`window.scroll(0,0)`

also I think this should be perhaps router middleware

app/elements/routing.html
```diff
@@ -17,19 +17,23 @@
     // More info: https://visionmedia.github.io/page.js/
     page('/', function () {
```
samccone: for instance 

```js
page('/', scrollToTop, function () {
...
});
```

app/scripts/app.js
chuckh: This was the suggested workaround from #202 and I have not test `window.scroll(0,0)`.

I created a function in app.js in case one might want to use it outside of the router.

app/elements/routing.html
chuckh: I tried to make easy for one disable scrolling to top. One might not always want to scroll to top for all routes, instead setting to last page position instead.

samccone: I agree :) you could omit the middleware per route with this approach :palm_tree:

addyosmani:
Change LGTM. Thanks @chuckh! 

I'll leave it open to us adding an entry to the FAQ about the middleware approach to this. I thought it was an interesting alternative :car:

carricha:
id mainContainer exists twice in PSK. Once in the drawer and once in main.
So how can document.getElementById('mainContainer') know which one to scroll to the top ?

Doesn't work for me.


<div id="mainContainer" class="style-scope paper-scroll-header-panel" style="top: 64px;">
      <paper-menu class="list x-scope paper-menu-0" attr-for-selected="data-route" role="menu" tabindex="0">

<div id="mainContainer" class="style-scope paper-scroll-header-panel" style="padding-top: 192px;">

Rgds

Issue #272, opened 2015-07-24
.paper-font-display1


gertcuykens:
https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/styles/app-theme.html#L143

For me its very confusing if your app theme is overwriting material design values from paper-styles.

```css
    .paper-font-display1 {
      font-size: 12px;
    }
```

Recommend to style something else to show case instead of overwriting recommended material design styles.

addyosmani:
Material Design does not define responsive typography at present which means that `recommended` in this context is unfortunately limited. What you get with paper-styles doesn't yet take responsiveness into account, meaning `paper-font-display1` will always mean the same thing regardless of context. 

We can either stay with our current approach or define a class that gets the job done but doesn't step over `paper-font-display1`. Interested in @chuckh's take on this.

chuckh:
My lean is to go with defining a class such as `greeting` and replace `paper-font-display1` with it.

All the places paper-font-display1` will need to be replaced to do so: https://github.com/PolymerElements/polymer-starter-kit/search?utf8=%E2%9C%93&q=paper-font-display1.

addyosmani:
I'm personally fine with us opting for a `greeting` class in place of `paper-font-display1`. Let's see if we can sneak this into 1.1

addyosmani:
Addressed by https://github.com/PolymerElements/polymer-starter-kit/issues/326

PR #326, opened 2015-08-24
Add page-title class for all page titles

chuckh:
Replaces using paper-font-display1 for page titles. Makes all page
titles have consistent formatting.  To get page-title to work with
`max-width: 600px` had to use `!important` for my-greeting-html.
My-greeting-html did not pick the page-title class formatting from
app-theme.html.

Any suggestions on handling this better are very welcome.

Fixes: #272


addyosmani:
Tested changes locally. LGTM. I would have approached the problem this way too btw.

addyosmani:
Thanks @chuckh!

arthurvr:
Probably better to squash commits in the future.

PR #327, opened 2015-08-24
Remove an accidentally added word

arthurvr:
This looked like unused and added by accident in https://github.com/PolymerElements/polymer-starter-kit/commit/98a7e3c37f824386b8b0028d6fb3014df0844f90.


samccone:
ha good spot :+1:

chuckh:
My bad. Thanks for catching that @arthurvr!

Issue #267, opened 2015-07-23
main.css vs app-theme.html


Frikki:
I am a bit confused about the purposes of these two files. I am aware that `app-theme.html` is a stylesheet shimmed with custom props by `custom style`, while `main.css` isn’t. But couldn’t every style block be in `app-theme.html`? Why have `main.css` unless for the purpose of critical style blocks that affect the page before elements are loaded, and is that its sole purpose?

addyosmani:
The purpose of `main.css` is to define styles that can be applied outside of Polymer's custom CSS properties implementation. Some of the use-cases include defining styles that you want to be applied for a splash screen, styles for your application 'shell' before it gets upgraded using Polymer or as you mentioned, critical style blocks that you want parsed before your elements are.

chuckh:
This subject is now covered in readme.md. See https://github.com/PolymerElements/polymer-starter-kit#application-theming--styling.

PR #328, opened 2015-08-24
Add shared styles

chuckh:
Add the new Polymer 1.1.0 share styles approach. Addresses #267.

### DIFF
--- a/app/elements/my-greeting/my-greeting.html
+++ b/app/elements/my-greeting/my-greeting.html
@@ -10,19 +10,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 <dom-module id="my-greeting">
   <template>
+    <style include="shared-styles"></style>
     <style>
       :host {
         display: block;
-      }
-
-      .page-title {
-        @apply(--paper-font-display2);
-      }
-
-      @media (max-width: 600px) {
-        .page-title {
-          font-size: 24px!important;
-        }
       }
     </style>
     <h2 class="page-title"><span>{{greeting}}</span></h2>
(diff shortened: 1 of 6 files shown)


app/elements/my-greeting/my-greeting.html
```diff
@@ -8,22 +8,14 @@
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 -->
 
+<link rel="import" href="../../styles/shared-styles.html">
```
moderndeveloperllc: Is this line needed as it's already being imported in the elements.html?

chuckh: Good catch. No it is not needed. I wil remove.

addyosmani:
Thanks for throwing this together, @chuckh. I've been reviewing https://github.com/Polymer/polymer/issues/1876 https://github.com/Polymer/polymer/issues/2229#issuecomment-134603556 &  https://github.com/Polymer/polymer/pull/2198 to catch up on the new approach. 

Your changes LGTM here, however testing locally with 1.1.1 I'm seeing some differences between Chrome and Safari (Mac):



I've verified that master otherwise appears to be fine:



Could you do another pass on testing the change out to see what might be causing the discrepancy?

chuckh:
@addyosmani did some more testing this morning. It worked for my on Chrome and Safari, but got the same results as you did but with Firefox instead.  I was using 1.1.1, but then I decide to try doing a bower update and that fixed it. Try doing bower update then try testing again.


### Results after bower update:

Safari (8.0.8 mac):
image

Firefox (40.0 mac): 
image

Chrome (44.0.2403 mac)
image

addyosmani:
LGTM

PR #333, opened 2015-08-27
Adding an ES2015 recipe.

nicolasgarnier:
Fixes #332 

@addyosmani PTAL

PR #335, opened 2015-08-27
Add starter element and gulp task el

chuckh:
You can create a new element by doing `gulp el —new my-element`.  This example will create a new folder under elements called `my-element ` and create a file called `my-element.html` in this folder.  `my-element` is based off `app/elements/starter-element.html`. 

Run `npm install` to install new dependency of minimist.

The purpose of this pr is to make it simpler to create a new basic element. I will update the readme.md with instructions if the team thinks we should add this. What do you think: @addyosmani, @arthurvr, @samccone?

### DIFF
--- a/app/elements/starter-element.html
+++ b/app/elements/starter-element.html
@@ -1,0 +1,40 @@
+<!--
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<dom-module id="starter-element">
+  <template>
+    <style include="shared-styles"></style>
+    <style>
+      :host {
+        display: block;
+      }
+    </style>
+    <paper-material elevation="1">
+      <h2 class="page-title">Welecome to <span>{{name}}</span>!</h2>
+      <p>This is your new element <span>{{name}}</span>.</p>
+    </paper-material>
+  </template>
+
+  <script>
+    (function() {
+      Polymer({
+        is: 'starter-element',
+
+        properties: {
+          name: {
+            type: String,
+            value: 'starter-element'
+          }
+        }
+      });
+    })();
+  </script>
+
+</dom-module>
--- a/gulp-tasks/el.js
+++ b/gulp-tasks/el.js
@@ -1,0 +1,24 @@
+'use strict';
+
+// Create new element from app/elements/starter-element.html
+// syntax: gulp el --new my-element
+module.exports = function ($, gulp, parseArgs, replace, fs) { return function () {
+  var newName = parseArgs(process.argv.slice(2)).new;
+  if (!fs.existsSync('app/elements/' + newName)){
+    if ( newName !== undefined && newName !== true && newName.indexOf("-") > 0) {
+      var folder = 'app/elements/' + newName;
+      var newFile = newName + '.html';
+      gulp.src('app/elements/starter-element.html')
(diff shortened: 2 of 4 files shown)


addyosmani:
Thanks a lot for working on this @chuckh. I'm a little torn on whether we should go in this direction. On one hand, I can see the utility value of just being able to run another task to create an element. On the other, this seems to overlap directly with scaffolding functionality we would normally tell folks to just use Yeoman for (via generator-polymer). My concern is setting a precedent for us adding in features we already have another tool for. Any thoughts? cc'ing in @robdodson for his take on this too.

arthurvr:
It's of course up to the team to decide this, but I feel like this really is reinventing the wheel. Yeoman does this perfectly and is created especially for things this. I also feel like Yeoman does it significantly better (the prompting, config, Yeoman generators are easily forkable, ...)

chuckh:
The reason I created this pr:

1. Since PSK has been release I have found myself not use yeoman anymore (used it a lot with Polymer < 1.0). ***One less tool to maintain***. Instead I just created a starter-element that I would manually copy when creating a new element. This just just automated the process some.
2. When I reviewed the PSK tutorial section to *Create a new element* that Kayce is working on, it seemed more complicated then needed. See https://raw.githubusercontent.com/kaycebasques/docs/psk/1.0/docs/start/psk.md Create a new element section
3. To show a method of using separate gulp task file, keeping gulpfile.js lighter. We might want to use this method for recipes like ***Adding an ES2015 recipe*** see https://github.com/PolymerElements/polymer-starter-kit/pull/333#discussion_r38106827, or deploying to GAE recipe see https://github.com/StartPolymer/polymer-starter-kit-plus/blob/master/gulp-tasks/deploy.js, or deploying to Firebase recipe see https://github.com/StartPolymer/polymer-starter-kit-plus/blob/master/gulp-tasks/deploy.js
4. To start a dialog about this approach

I am fine with whatever outcome the team thinks is best:  include in main PSK, making it a recipe, or not using it at all.

arthurvr:
> One less tool to maintain.

If you mean something like this can replace Yeoman I think you're forgetting quite a bit of Yeoman's features and usecases. Just to give a few super easy examples, the Yeoman generator fills in [dependencies](https://github.com/yeoman/generator-polymer/blob/master/el/templates/element.html#L12) and the [element name](https://github.com/yeoman/generator-polymer/blob/master/el/templates/element.html#L28). It even has a GUI app.

Besides from that, I think the code inside of `gulp-tasks/el.js` shouldn't live in one's project but in the Yeoman generator (or in their fork if that's needed). Separation of concerns, right? Things like this are usable in so many projects that I don't see the point of repeating it in many projects, it's better to use already existing tooling (like Yeoman) therefore. 

> Instead I just created a starter-element that I would manually copy when creating a new element. This just just automated the process some.

That automation is exactly what Yeoman does and is meant to do. I still don't see why we'd reinvent the wheel. 

> To show a method of using separate gulp task file, keeping gulpfile.js lighter.

:+1: Although, our gulpfile is still relatively small so I personally don't think that's needed yet.

JosefJezek:
We have snippets, check out the link https://github.com/robdodson/Atom-PolymerSnippets

robdodson:
The Polymer generator should recognize the PSK structure and allow you to quickly create an element using [the `el` subgenerator](https://github.com/yeoman/generator-polymer#element-alias-el). In the latest release it should also scaffold out a test for you if you want it. My (totally biased vote :smile_cat:) is to keep this kind of stuff in a generator and keep the gulpfile as small as possible.

chuckh:
Thanks everyone for your consideration and thoughtful feedback. I agree this is better handled by Yeoman Generator for Polymer. So I'm going to close this pr. 

We may want to consider updating the readme.md at https://github.com/PolymerElements/polymer-starter-kit/#yeoman-support suggesting to use Yeoman Generator for Polymer as the recommended way to get and work with PSK.

@robdodson nice working on adding the `tests` and `--docs` to the generator.

robdodson:
ah i can't take credit for those, they came in as PRs from the community.
Open source FTW! :)

PR #343, opened 2015-09-01
Using local gulp and bower, simplified setup with npm postinstall script.

uldissturms:
This removes the need of global gulp and bower thus simplifies the setup, one command will install all node modules and as part of post-install script - bower components. Hope this helps.
```sh
npm install
```


uldissturms:
I signed it!

addyosmani:
Thanks for your PR. We've previously discussed the pros and cons of opting for a similar approach and decided that given PSK's (predominantly beginner) audience, exposure to the direct tools being used vs. hiding them behind npm scripts makes the most sense for our users. Sorry :dizzy:

PR #349, opened 2015-09-08
Implement support for Chrome for Android splashscreen

addyosmani:
reviewer: @mounirlamouri
Validated using http://mounirlamouri.github.io/manifest-validator/


samccone:
lgtm :+1:

mounirlamouri:
lgtm

mounirlamouri:
Though, you might want bigger icons for the splashscreen, on high resolution devices, the ones currently defined in the Manifest might not be enough.

addyosmani:
@mounirlamouri is there a good device to validate resolution appropriateness on? Checked against an N5 but can test N6 and N9.

mounirlamouri:
The N6 is 3x or 3.5x I believe so it will ideally use a 128*3 or 128*3.5 icon for the splash screen. We allow ourselves to pick a slightly smaller one if available (2x in that case so 256px) but I don't think you are providing one.

addyosmani:
@mounirlamouri Added a 3x (384x384) entry. PTAL.

mounirlamouri:
Hmm, I now see why you were confused with the sizes, GitHub formatted my ```128x3``` to something odd. Sorry about that, I should have double checked.

Otherwise, LGTM. It would be better with a 256x256 but Chrome is able to take the bigger one and resize it if needed.

addyosmani:
Thanks for the reviews, both! Landing :pizza:

PR #353, opened 2015-09-09
🍭 Fix jshint strict warning

### LINKED ISSUE #352: Enable use of use strict in elements
Currently `use strict` in custom elements will cause jshint to fail. Will look for solution, possibly using http://jshint.com/docs/options/#nonew

samccone:
ref https://github.com/jshint/jshint/issues/438

fixes #352


chuckh:
LGTM. Nice work @samccone.

PR #354, opened 2015-09-10
Add use strict to elements

chuckh:
Adds `use strict` to my-greeting.html and my-list.html now that jshint
config is fixed to handle this.


samccone:
Lgtm 👏👌

PR #356, opened 2015-09-13
Move bower_components into app folder

MartinsThiago:
When we execute `bower install` from command line, dependencies are placed as the image below

image

I propose placing `bower_components` inside app folder because it's where dependencies are coming from after build... while developing IDEs can provide better IntelliSense once it can find references

image

## Warning
Sorry for creating a pull request not 100% finished... I got it to work with simple `gulp serve`, but running a complete build... webcomponents.js appear two times in the final script and it causes the app not to start

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,1 @@
+{"directory": "app/bower_components"}
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 node_modules
 dist
-bower_components
+app/bower_components
 .tmp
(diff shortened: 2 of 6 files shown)


arthurvr:
I'd prefer to stick to the current structure, for the reasons outlined in the [blogpost](http://yeoman.io/blog/bower_components-in-project-root.html) of the Yeoman team.

addyosmani:
I'm personally :+1: on maintaining the current structure for the same reasons outlined above by the Yeoman team.

MartinsThiago:
I read the post, and yup, it's a good reason... of course with some upsides and downsides, but anyway, it's being great to work with Polymer

:smiley:

PR #364, opened 2015-09-17
Make sure that index.build.html is deleted before proceeding

timeu:
`del()` is asynchronous which can cause problems if there are other tasks (i.e. deploy to gh-pages) that depend on the `default` task.
The task will pick up `index.build.html` and thus fail.  
This change will make sure that the `rename-index` task only signals completion once `index.build.html` is deleted.


sindresorhus:
This isn't correct either as `gulp.src('dist/index.build.html')` will race with the `del` call. `del` should be executed after the gulp stuff.

Solution: https://github.com/gulpjs/gulp/blob/master/docs/recipes/delete-files-folder.md#delete-files-in-a-pipeline

samccone:
hey @timeu thanks for getting the ball rolling on this one, I have migrated the issue / PR :)

timeu:
@samccone Great thanks !

PR #365, opened 2015-09-17
Fix 'clean' task in gulpfile.js

frncesc:
Task `clean` does not return, thus breaking default gulp task.
Added a `return` statement to fix it.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -200,7 +200,7 @@ gulp.task('cache-config', function (callback) {
 
 // Clean output directory
 gulp.task('clean', function (cb) {
-  del(['.tmp', 'dist'], cb);
+  return del(['.tmp', 'dist'], cb);
 });
 
 // Watch files for changes & reload


samccone:
lgtm :+1:

arthurvr:
Are we running into any concrete issues? Generally a task not returning anything should work, or am I wrong? We do use the `cb` argument. We have other tasks too that don't return anything.

arthurvr:
Dupe of #364 btw.

samccone:
> If you do not return a stream, then the asynchronous result of each task will not be awaited by its caller, nor any dependent tasks.

http://stackoverflow.com/questions/26079118/does-a-gulp-task-have-to-return-anything

So by break, it could (and does) introduce racey cases ref #364 which everyone :heart: loves.

:+1: to ship this :sparkles:

arthurvr:
Oh, alright. :+1:

arthurvr:
(this one should be closed as a dupe though)

samccone:
personally I prefer the return approach vs the cb approach

frncesc:
I signed it!



On Thu, Sep 17, 2015 at 2:20 PM, googlebot <notifications@github.com> wrote:

> Thanks for your pull request. It looks like this may be your first
> contribution to a Google open source project, in which case you'll need to
> sign a Contributor License Agreement (CLA).
>
> [image: :memo:] *Please visit https://cla.developers.google.com/
> <https://cla.developers.google.com/> to sign.*
>
> Once you've signed, please reply here (e.g. I signed it!) and we'll
> verify. Thanks.
> ------------------------------
>
>    - If you've already signed a CLA, it's possible we don't have your
>    GitHub username or you're using a different email address. Check your
>    existing CLA data <https://cla.developers.google.com/clas> and verify
>    that your email is set on your git commits
>    <https://help.github.com/articles/setting-your-email-in-git/>.
>    - If you signed the CLA as a corporation, please let us know the
>    company's name.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/PolymerElements/polymer-starter-kit/pull/365#issuecomment-141061864>
> .
>



-- 
Francesc Busquets
francesc@gmail.com

samccone:
Ha no need for both a cb and return 😎

addyosmani:
@arthurvr let's make a call on this vs the other PR. Does 364 lgty as is too? Seems fine to me. Ultimately comes down to cb vs. return preferences :lipstick:

samccone:
I'm 👍 for using a return since del already returns a stream and gulp "just works" with that interface

robdodson:
+1 for returning. Seems like the nicer approach

arthurvr:
:+1:

sindresorhus:
This is incorrect. del@1 only had a callback interface and returning here does nothing. You now return *and* use the `cb` interface.

You should rather upgrade to del@2 (which has a promise interface) and just return instead of the `cb`.

addyosmani:
@sindresorhus Reverted in https://github.com/PolymerElements/polymer-starter-kit/pull/377. I wasn't aware del@2 supports a promise interface. Let's switch up to that instead.

samccone:
Word was looking at the docs for del2 😰😭

Yeah https://github.com/gulpjs/gulp/blob/master/docs/API.md#async-task-support then a return will just work 👏 good catch @sindresorhus

PR #368, opened 2015-09-17
Dry up router scrollToTop code

robdodson:


app/elements/routing.html
```diff
@@ -15,25 +15,26 @@
     // We use Page.js for routing. This is a Micro
     // client-side router inspired by the Express router
     // More info: https://visionmedia.github.io/page.js/
+    page('*', function (ctx, next) {
```
arthurvr: `function()` would be consistent with line 13, while line 23 uses something else. What's the correct version?

robdodson: elsewhere, like in app.js, we do `function()` so maybe they should all be changed to that?

arthurvr: > so maybe they should all be changed to that?

:+1:

chuckh:
@robdodson we decided to leave `app.ScrollToTop()' for each route so users could choose which routes they wanted this behavior. We did not want to force every route to have this behavior. On some routes one might want to return to last position instead top for example.

robdodson:
I see. Can you link me to the discussion? It kinda feels like it should be the other way around. By default the code could be DRY and if the user wants to exempt a route they could check which route it is in the `*` handler and just call next and return

samccone:
page.js has middleware (just like express) which I much prefer to the `*` approach, it is much more explicit and opt in

for instance:

```js
page('/foo', scrollToTop, function(){...})
page('/zap', function(){...})
```

robdodson:
good call, let's do that instead

robdodson:
Rebased. @samccone @chuckh @arthurvr wdyt?

chuckh:
LGTM!

samccone:
:+1: from me.. in it goes! :tada::tada::tada::tada::tada:

PR #385, opened 2015-09-24
🐐 Bring more inline with gulp best practices

### LINKED ISSUE #289: Gulp tasks should be named functions wrapped and without runSequence
The gulp tasks have been written against gulp convention.  The true intention of the gulp design is to write regular Javascript functions, and only make the task for calling them. With this format, you would never need `runSequence` which uses(or used to use) `gulp.start()`. 

If you need sync, use a promise library like [blue bird performance promise library](https://github.com/petkaantonov/bluebird/blob/master/API.md). I believe Gulp 4.0 has it's own promise's but I prefer performance bluebird. 

This is from the authors of Gulp. Sources:
[gulpjs/gulp#755](https://github.com/gulpjs/gulp/issues/755#issuecomment-61884019)
[gulpjs/gulp#458](https://github.com/gulpjs/gulp/issues/458#issuecomment-62057370)
example on proper convention:

```javascript
function getJsFiles() {
    var sourcePaths = [
        './app/scripts/**/*.js',
        '!./app/scripts/**/*.spec.js',
        '!./app/scripts/app.js'
    ];

    var sources = gulp.src(sourcePaths, { read: false }).pipe(angularFilesort());

    return gulp.src('./app/index.html')
        .pipe(injector(sources, { ignorePath: 'app', addRootSlash: false }))
        .pipe(gulp.dest('./app'));
} 
```
----------------------------
```javascript 
gulp.task('js', function () {
    jsStream = getJsFiles();
});
```

samccone:
Extract non-specific tasks to fns to be more inline with "the gulp way✨"

Fixes #289

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -49,6 +49,49 @@ var styleTask = function (stylesPath, srcs) {
     .pipe($.size({title: stylesPath}));
 };
 
+var jshintTask = function (src) {
+  return gulp.src(src)
+    .pipe($.jshint.extract()) // Extract JS from .html files
+    .pipe($.jshint())
+    .pipe($.jshint.reporter('jshint-stylish'))
+    .pipe($.if(!browserSync.active, $.jshint.reporter('fail')));
+};
+
+var imageOptimizeTask = function (src, dest) {
+  return gulp.src(src)
+    .pipe($.cache($.imagemin({
+      progressive: true,
+      interlaced: true
+    })))
+    .pipe(gulp.dest(dest))
+    .pipe($.size({title: 'images'}));
+};
+
+var optimizeHtmlTask = function (src, dest) {
+  var assets = $.useref.assets({searchPath: ['.tmp', 'app', 'dist']});
+
+  return gulp.src(src)
+    // Replace path for vulcanized assets
+    .pipe($.if('*.html', $.replace('elements/elements.html', 'elements/elements.vulcanized.html')))
+    .pipe(assets)
+    // Concatenate and minify JavaScript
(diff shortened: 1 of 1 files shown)


addyosmani:
Reviewed.

samccone:
updated!

gulpfile.js
```diff
@@ -49,6 +49,50 @@ var styleTask = function (stylesPath, srcs) {
     .pipe($.size({title: stylesPath}));
 };
 
+var jshintTask = function (src) {
+  return gulp.src(src)
+    .pipe(reload({stream: true, once: true}))
```
didrocks: Hey, it seems to me that you are reintroducing the double reload that was removed as per PR #370, is that intended?

samccone: I just need to rebase and it should go away (unless Git gets really confused)👌 good spot though I will doubly make sure it gets removed!

samccone:
yep it went away @didrocks :)

didrocks:
@samccone great! thanks for confirming ;)

addyosmani:
Tried running locally again. Looks like we're now breaking on the HTML task and are running into some linting issues:

```
polymer-starter-kit/gulpfile.js
  line 58  col 2   Missing semicolon.
  line 68  col 2   Missing semicolon.
  line 93  col 2   Missing semicolon.
  line 91  col 21  'dest' is not defined.

  ⚠  4 warnings
```

samccone:
should be GTG now :checkered_flag:

addyosmani:
Tested again and everything is passing now. Have some :cake:!

Issue #283, opened 2015-07-28
Mixins applied to an element from an external stylesheet are missing in builds


sfeast:
Here's one way to reproduce:

1) Define a custom mixin in app-theme.html in the :root selector. Something like:
```
--demo: {
      color:yellow;
      background-color:blue;
    };
```
2) Move the styles in my-greeting.html to an external stylesheet (such as my-greeting.css).
3) Apply the mixin as described here - https://www.polymer-project.org/1.0/docs/devguide/styling.html#custom-css-mixins to your my-greeting.css. For example you can do:
```
:host {
	display: block;
	@apply(--demo);
}
```

Result:
When using ```gulp serve``` it works as expected, but when building the app (i.e. build with ```gulp```), the resulting app does not include the mixin styling.

Notes:
1) Custom properties are working with the build, just not mixins.
2) I noticed this when the ```iron-flex-layout ``` mixins I had applied were not appearing in my built app.
3) Mixins are working when applied in the element's html under a ```<style>``` tag, this seems to just be in external stylesheets.

devAtPolydeals:
@sfeast  Does it work from the app directory? I doubt external sytlesheets will work with mixins, just because css doesn't know about mixins and I don't know if Polymer will flatten the css before it renders the page. You can try making it an HTML file and load Polymer and CSS in that HTML file, I am guessing this might work.

sfeast:
@devAtPolydeals I think you're asking me if mixins work when served (```gulp serve``` for example) & if so the answer is yes they do work in that case.

The problem seems to be that some step in the build process is filtering out the @apply statements from the css since they are missing in elements.vulcanized.html (just using the default build process here). If I manually add them back in then things work as expected.

robdodson:
@azakus have you seen this with vulcanize?

sfeast:
FYI - the culprit seems to be cssmin. I'm not yet very familiar with the details of the default build setup but if you remove

```.pipe($.if('*.css', $.cssmin()))```

from the 'styles' task in gulpfile.js, then although the css does not get minified, it does have the mixins left intact.

Also note that by default cssmin is not applied against any included Polymer elements with external css, so those elements come out ok.

robdodson:
Hm darn.. that make sense.

@azakus any idea how we could minify <style is="custom-style"> elements appearing in external imports?

azakus:
@robdodson No, I've yet to find a good way to minimize CSS that doesn't blow stuff up.

[...]

PR #386, opened 2015-09-24
⭐️ Support external CSS/JS out of the box.

### LINKED ISSUE #225: vulcanize task fails silently after new elements added in ./app/elements
to post an issue that i had extending the project, adding my own elements having child 'js' folders with js there instead of in-line in the element.

When you take the project and make this kind of addition, you expect existing gulp:copy task to do its job and to prepare for following vulcanize step to just work with the new assets under ./app/elements.

What actually happens is that vulcanize fails silently because the new element gets copied over to ./dist/elements WITHOUT its children. Vulcanize will require that "my-newelement/js/mynewelement.js" exists over in the copy tasks dest folder.

Well , until a developer makes the following change to "copy" task, ( gulp.js#97 ) it all fails in a way where there is not any good log...

OLD gulp.js:

  var elements = gulp.src(['app/elements/**/*.html'])
    .pipe(gulp.dest('dist/elements'));

New gulp.js after added a new element:

  var elements = gulp.src(['app/elements/**/*.html', 'app/elements/**/*.js'])
    .pipe(gulp.dest('dist/elements'));

Im a newby in Polymer but, it took a lot of digging in order to find out that 'vulcanize' task actually needed the 'js' folder for the new element to be over in ./dist/elements/my-newelement/js

samccone:
Fixes #225

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -98,7 +98,9 @@ gulp.task('copy', function () {
     'bower_components/**/*'
   ]).pipe(gulp.dest('dist/bower_components'));
 
-  var elements = gulp.src(['app/elements/**/*.html'])
+  var elements = gulp.src(['app/elements/**/*.html',
+                           'app/elements/**/*.css',
+                           'app/elements/**/*.js'])
     .pipe(gulp.dest('dist/elements'));
 
   var swBootstrap = gulp.src(['bower_components/platinum-sw/bootstrap/*.js'])


alexpaluzzi:
This is needed. It's the first thing I manually change every time I start a new project. Thanks @samccone

addyosmani:
Changes LGTM. Thanks for getting this together so quickly, @samccone! :pizza:

samccone:
This fixes #283

Issue #373, opened 2015-09-20
gulp Error: ENOENT, open '.../dist/index.build.html'


vl4dt:
I just wanted to try the latest PSK, clonned it and installed dependencies as usual.

Tried many things, cleared npm cache, updated node, etc., keep getting the following error running gulp:

```bash
(master) psk110 ツ gulp
[01:54:48] Using gulpfile ~/Projects/psk110/gulpfile.js
[01:54:48] Starting 'clean'...
[01:54:48] Finished 'clean' after 246 ms
[01:54:48] Starting 'default'...
[01:54:48] Starting 'copy'...
[01:54:49] Starting 'styles'...
[01:54:51] styles all files 98 B
[01:54:51] Finished 'styles' after 1.2 s
[01:54:52] copy all files 10.97 MB
[01:54:52] Finished 'copy' after 3.7 s
[01:54:52] Starting 'elements'...
[01:54:52] Finished 'elements' after 6.62 ms
[01:54:52] Starting 'jshint'...
[01:54:53] Starting 'images'...
[01:54:54] Starting 'fonts'...
[01:54:54] Starting 'html'...
[01:54:55] Finished 'fonts' after 880 ms
[01:54:56] Finished 'jshint' after 4.49 s
[01:54:56] images all files 35.41 kB
[01:54:56] Finished 'images' after 3.1 s
[01:54:56] html all files 51.25 kB
[01:54:56] Finished 'html' after 2.17 s
[01:54:56] Starting 'vulcanize'...
[01:55:01] Finished 'vulcanize' after 4.54 s
[01:55:01] Starting 'rename-index'...
[01:55:01] Finished 'rename-index' after 2.35 ms
[01:55:01] Finished 'default' after 13 s
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open '/Users/vl4dt/Projects/psk110/dist/index.build.html'
    at Error (native)
(master) psk110 ✖
```

flaiker:
I just did the following:
```
$ git clone git@github.com:PolymerElements/polymer-starter-kit.git
$ cd polymer-starter-kit
$ npm install -g gulp bower && npm install && bower install
$ gulp
```

And it worked fine for me. Can you try installing the dependencies again (sometimes they timeout and some are missing) and also renaming your project folder to `polymer-starter-kit` (see #257).

[...]

PR #390, opened 2015-09-25
del requires the parent directory to be passed as the second argument

vl4dt:
This is the final fix to the issue:
[#373](https://github.com/PolymerElements/polymer-starter-kit/issues/373)

Looks like [**del**](https://www.npmjs.com/package/del) is not able to find the file for some reason in my system, does not return a promise, it returns **_undefined_** if the parent directory is not provided.

I can provide any additional details you consider may help to identify the origin of the issue if for some reason something in my setup causes [**del**](https://www.npmjs.com/package/del) to behave differently.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -166,7 +166,7 @@ gulp.task('rename-index', function () {
   gulp.src('dist/index.build.html')
     .pipe($.rename('index.html'))
     .pipe(gulp.dest('dist/'));
-  return del(['dist/index.build.html']);
+  return del(['dist/index.build.html'], gulp.src('dist'));
 });
 
 // Generate config data for the <sw-precache-cache> element.


vl4dt:
I signed it!

samccone:
So I am pretty confused how this patch makes things work for you  :horse: 

Let's take a trip down what this does

opts is being passed directly to del

and then opts is passed here :eyes: 
https://github.com/sindresorhus/del/blob/v1.1.1/index.js#L34

which is then passed into globby
https://github.com/sindresorhus/globby/blob/v1.2.0/index.js#L59

where is is passed into `setIgnore`
https://github.com/sindresorhus/globby/blob/v1.2.0/index.js#L31

where is should do nothing...

any thoughts :)

samccone:
ping @vl4dt  :)

vl4dt:
hey there :)
is taking me some time to find out why is different in my system mostly for time constraints but, I'm on it, been doing some debugging in this machine and will do in some other linux box I use at work where the issue this patch fixes doesn't occur, but want to see what happens internally, will keep you posted

samccone:
:+1: sgtm keep me posted and thanks a bunch!

samccone:
any progress or info @vl4dt ?

I am going to close this for now, but we can reopen if we can reproduce

vl4dt:
well after all I was unable to trace the reason why that worked, however now that I updated the OS, XCode and PSK everything works fine so, this is no longer needed, sorry for the delayed response but has been a very busy month, in the end I'm happy it works perfect :)

thanks guys

Issue #369, opened 2015-09-17
Polyclean fails after adding ES2015 code


alexpaluzzi:
I wasn't sure whether to drop this in the `polyclean` repo or this one--

I had this working the other day, so I started from scratch but got the same error. Going to continue to investigate on my end, but wanted to create this in the meantime in-case someone already knows about it.

```
[16:02:14] Starting 'clean'...
[16:02:14] Finished 'clean' after 269 ms
[16:02:14] Starting 'default'...
[16:02:14] Starting 'copy'...
[16:02:14] Starting 'styles'...
[16:02:15] styles all files 98 B
[16:02:15] Finished 'styles' after 1.18 s
[16:02:17] copy all files 8.82 MB
[16:02:17] Finished 'copy' after 2.94 s
[16:02:17] Starting 'elements'...
[16:02:17] Starting 'js'...
[16:02:18] Finished 'elements' after 1.51 s
[16:02:19] Finished 'js' after 1.77 s
[16:02:19] Starting 'jshint'...
[16:02:19] Starting 'images'...
[16:02:20] Starting 'fonts'...
[16:02:20] Starting 'html'...
[16:02:21] Finished 'images' after 2.21 s
[16:02:21] Finished 'fonts' after 1.1 s
[16:02:23] Finished 'jshint' after 3.85 s
[16:02:23] html all files 54.89 kB
[16:02:23] Finished 'html' after 2.29 s
[16:02:23] Starting 'vulcanize'...
[16:02:27] Error in plugin "polyclean"
Message:
    Unexpected token: punc ())
```

Node 4.0.0
Windows 7
(Will be on Linux machine tonight and plan on testing there.)

alexpaluzzi:
Everything runs until you start actually changing stuff to ES2015 code.

Following: http://www.code-labs.io/codelabs/polymer-es2015/
The `Unexpected token: punc ())` occurs after I added all of the fat arrows to the `routing.html` file.
When I took them all out, it ran again. Then I tried changing the `my-list` component to use a ES2015 class. Now the error from polyclean is: `Unexpected token: name (MyList)`.

I think something is out of order here. I will continue to work on it.

alexpaluzzi:
Seem to have found the issue:
https://github.com/mishoo/UglifyJS2/issues/448

This is probably worth mentioning in the ES2015 doc.

alexpaluzzi:
[...]

PR #421, opened 2015-10-07
Revert back to using Vulcanize over Polybuild

addyosmani:
Per the discussion on our group thread with the team, reverting back to Vulcanize for now.

Polybuild and the transitive dependencies it has currently break in a number of different scenarios, spanning breakage of apps using Firebase in [polyclean](https://github.com/PolymerLabs/polyclean/issues/4) through to issues with how it treats Service Worker scripts, [ES2015](https://github.com/PolymerElements/polymer-starter-kit/issues/369) and so forth. We're confident these issues can be resolved upstream but wish to restore build stability to PSK using the same setup we did in previous releases. 

r @samccone @chuckh


samccone:
:+1: tested it all locally :ship: :goat:

PR #424, opened 2015-10-08
optimizeHtmlTask

gertcuykens:
Still think this should be made more pluggable using the tasks directory so you don't have to mess with the gulpfile itself.

### DIFF
--- a/docs/add-es2015-support-babel.md
+++ b/docs/add-es2015-support-babel.md
@@ -66,16 +66,15 @@ gulp.task('default', ['clean'], function (cb) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


samccone:
good catch, thanks @gertcuykens 

per 
> Still think this should be made more pluggable using the tasks directory so you don't have to mess with the gulpfile itself.

That does sound super good, would you be interested in taking a stab?

gertcuykens:
Can you --ammend this also :D

`gulp.watch(['app/{scripts,elements}/**/*.js'], ['jshint', 'js']);`
to
`gulp.watch(['app/{scripts,elements}/**/{*.js,*.html}'], ['jshint', 'js']);`

PR #426, opened 2015-10-09
PSK + more-routing (and app-element centric vs dom-bind centric)

ktiedt:
I'm not expecting this to get merged to master, GitHub does not appear to provide a means to suggest a PR become a new branch... so quick run down of what I did here.

1) I moved practically everything into custom elements (psk-*) and moved app.js into basically an empty shell.This included moving the CSS around quite a bit and tweaking some element IDs to better work with this.$ references fo replace the app.js code.

2) Replaced Page.js with more-routing as it seemed to be the most ready router at the time and I needed some practice before moving forward with a work project using Polymer + Routing (other than Page.js)

I intended to provide this as a means of showing the PSK with another router AND using an <app-element> to cover all the top level bindings etc.


ikeagold:
@ktiedt :+1: , its interesting to try PSK on only Polymer components and also i dont like 'dom-bind' centric too :cyclone:

addyosmani:
@ktiedt Thanks for opening up this PR for discussion. We've been talking with the rest of the Polymer team about what the best set of idiomatic architecture patterns for PSK might look like lately, also factoring in concerns such as SEO and progressive enhancement.

The first thing we can improve (pattern-wise) here is as you pointed out - moving away from a `dom-bind` heavy architecture to one that just uses elements all the way down. I actually wouldn't be surprised if this is the direction we decided to go down for one of the next major releases - it's certainly more readable than the current state of our index.html. 

I do think we need to discuss what implications to PE this particular pattern has and whether that's a concern for the project. Let's keep this open to track notes and perhaps we can evolve this PR to update the architecture or follow it up in the future to do so?

On `more-routing`: we're planning on deprecating this element in the near future as it's always been something of an experiment. We are planning on replacing it with something far better, but it's not quite ready just yet. I'm personally otherwise a fan of the new routing work by @atotic over at https://github.com/atotic/excess-router and you might find it a suitable replacement in the short term if you prefer something more declarative than page.js. 

To avoid forcing users to keep switching routing solutions, I'd be tempted to say we should stick with page.js for the short term until the official routing story is more fleshed out but could do something like write a recipe in our docs folder for how to swap it for `excess-router`.

addyosmani:
Also looping in @robdodson @samccone @chuckh into the mix.

ktiedt:
@addyosmani Absolutely, as I stated, I had no expectations of this 'landing', unless it was a separate branch that users could swap to just to see another approach. I am a bit bummed that more-routing is expected to go away.... I did see there is a relatively new branch for 'maybe-router' and thought maybe that was the way forward... 

Would it be possible to get more solid consensus on routing sooner than later? This is something that is hobbling lots users trying to sort through all the early routing options and wanting to stay with the 'official' (if there is one) or at least the solution that gets a nod from the Polymer Team.

samccone:
based on your more recent research @ebidel Is this still a pattern you think we should be using?

samccone:
@ktiedt I am going to close this, but use this issue as a reference point (plz do not delete your remote branch)

this work will be coming into play soon :)

ktiedt:
Sure thing, thanks for the heads up @samccone

samccone:
thank you @ktiedt for the work and the investigation you have done here, it is invaluable :clap:

abdonrd:
Nice work, @ktiedt!

I am also very interested about the new routing approach.

ikeagold:
@ktiedt can you help me, 1: `:host ::content paper-material` or 2: `:host::content paper-material`? thanks :cold_sweat:

Compufreak345:
I also would be really pleased to see any news on the new routing system. For now I will look into migrating away from more-routing (that worked quite OK with some workarounds, but has many unresolved issues) to probably excess-router or page.js.

abdonrd:
I created https://github.com/PolymerElements/polymer-starter-kit/issues/576 to discuss about 'dom-bind centric' vs 'app-element centric'.

Issue #379, opened 2015-09-23
When disabling shady DOM PSK stops working


David-Mulder:
In app.js

    document.getElementById('mainContainer')

can't be found when shadom DOM is used and thus the entire PSK stops working.

davidrissato:
is this element in your main html and outside a template is=dom-bind? Shadow DOM intends to hide things, and if this tag isn't in the main html, it won't be found under document's content children DOM. It will be hidden in the shadow root.

David-Mulder:
Yeah I know... it's not a hard thing to fix... it's just that by default it should support the non-polyfilled Polymer as well.

addyosmani:
Another comment from #438 cross-posting:

the following logic in `app.js` makes an assumption by accessing internal implementation specific logic, namely an internal id

```javascript
  // Scroll page to top and expand header
  app.scrollPageToTop = function() {
    document.getElementById('mainContainer').scrollTop = 0;
  };
```

wouldn't it be better to implement this as follows?

```javascript
  // Scroll page to top and expand header
  app.scrollPageToTop = function() {
    document.getElementById('mainArea').scroller.scrollTop = 0;
  };
```

by adding the `id="mainArea"`, *or something similar*, in `index.html`

```html
.
.
.
    </paper-scroll-header-panel>

    <!-- Main Area -->
    <paper-scroll-header-panel id="mainArea" main condenses keep-condensed-header>

    <!-- Main Toolbar -->
.
.
.
```

Issue #178, opened 2015-06-16
Improve Performance


JosefJezek:
- [ ] Use inline styles to reserve app “regions”.
- [ ] Vulcanize and crush the element bundles (excluding Polymer itself).
- [ ] Conditionally and async load the Web Components lite polyfill.
- [ ] Async load the element bundles. (This will also set up a race for Polymer itself).

https://aerotwist.com/blog/polymer-for-the-performance-obsessed/

moderndeveloperllc:
Aren't one and two already being done with gulp? I for one _like_ to have all my elements loaded in one vulcanized page. Maybe I'm in the minority on that though.

kenchris:
I think that #1 is interesting because it might not be so obvious for people how to do that in a nice way

addyosmani:
Today, @ebidel showed us a different take on the lazy loading pattern from @paullewis that was based on ES6 promises. I'd be up for us getting behind a single solid pattern we can bake into PSK as a default.

addyosmani:
> Async load the element bundles. (This will also set up a race for Polymer itself).

@wibblymat has been doing research into the state of `async` friendliness of Polymer elements this past week. I don't think we're quite at the point where we can include opinion on it here, but this should be possible for it in the not too distant future.

>  Conditionally and async load the Web Components lite polyfill.

See above.

> Vulcanize and crush the element bundles

We're exploring what better support for element bundles would look like in Vulcanize. I'm told this is WIP.

> Use inline styles to reserve app “regions”.

It would be trivial for us to set this up for the default paper-* based layout, but I'm wondering how useful baking this in will practically be. Every app is going to be different. Some folks will prefer to implement their own splash screens, others will prefer rendering an application scaffold. Let's keep thinking about the best way to integrate support for this.

paullewis:
I have a tweak to my pattern that may bring it in-line with @ebidel's, but without some of its quirks. I reckon next week will get us there. We are close.

addyosmani:
> I have a tweak to my pattern that may bring it in-line with @ebidel's, but without some of its quirks. I reckon next week will get us there.

:rocket:

bppffnn:
JSperf is broken at the moment because @mathiasbynens so you can't test performance!!

addyosmani:
We're going to start work on integrating performance patterns from 1.0.4 onwards, starting with async loading the Web Component polyfills. We may look at async loading the element bundles after that.

[...]

PR #431, opened 2015-10-10
Use Polymer.dom in app.js. Fixes Gh-379

robdodson:
This PR switches `app.js` over to using Polymer's Shady DOM API for querying the document (https://www.polymer-project.org/1.0/docs/devguide/local-dom.html#dom-api). This should preserve Shadow boundaries. I believe moving forward if you're querying the DOM you'll always want to use  `Polymer.dom` methods.

This should fix https://github.com/PolymerElements/polymer-starter-kit/issues/379 as it's now querying the Light DOM `paper-scroll-header-panel` and using its API to scroll, instead of the `mainContainer` element which should be hidden by Shadow DOM.

@samccone @chuckh @addyosmani PTAL

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -17,8 +17,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   app.displayInstalledToast = function() {
     // Check to make sure caching is actually enabled—it won't be in the dev environment.
-    if (!document.querySelector('platinum-sw-cache').disabled) {
-      document.querySelector('#caching-complete').show();
+    if (!Polymer.dom(document).querySelector('platinum-sw-cache').disabled) {
+      Polymer.dom(document).querySelector('#caching-complete').show();
     }
   };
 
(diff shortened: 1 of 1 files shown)


chuckh:
I reviewed looks good. Though I did not get a chance to fully test. 

P.S. sorry I have be quite lately, been doing a crazy 3 way move (of home, office and family). Should be over the hump later this week.

robdodson:
No worries @chuckh, thanks for the review :)

addyosmani:
Thanks for the PR, @robdodson!

Seeing some issues in Safari:

screen shot 2015-10-12 at 10 59 50

and Firefox without the WC flags on:

screen shot 2015-10-12 at 11 03 18

gertcuykens:
Can you also add this webComponentsSupported please from
https://youtu.be/Yr84DpNaMfk?t=8m48s

```
  var webComponentsSupported = (
    'registerElement' in document &&
    'import' in document.createElement('link') &&
    'content' in document.createElement('template')
  );

  if (!webComponentsSupported) {
    var script = document.createElement('script');
    script.async = true;
    script.src = 'bower_components/webcomponentsjs/webcomponents-lite.js';
    script.onload = finishLazyLoading;
    document.head.appendChild(script);
  } else {
    finishLazyLoading();
  }
  
  function finishLazyLoading() {
    window.Polymer = window.Polymer || {};
    window.Polymer.dom = 'shadow';
  }
```

ebidel:
@gertcuykens that is unrelated to this fix. https://github.com/PolymerElements/polymer-starter-kit/issues/178 are the droids you're looking for.

addyosmani:
Friendly ping @robdodson whenever you get a chance to look at this again :)

robdodson:
@addyosmani ah I see the problem. It's here: https://github.com/PolymerElements/polymer-starter-kit/pull/431/files#diff-685dfa901a119f6590e773891ffd6052R15

Polymer isn't loaded yet on Safari or FF because the imports are async. I think the two fixes would be to either move everything inside WebComponentsReady callback or don't use the `doc` variable and spinkle `Polymer.dom` calls everywhere which doesn't feel very DRY to me. wdyt?

robdodson:
Actually nvm, if I move everything inside WebComponentsReady, then the opposite problem occurs, it's now broken on Chrome which is looking for the methods on `app` immediately

robdodson:
OK I reverted to just using Polymer.dom throughout. A quick check in Chrome, FF, and Safari seemed to look ok. @addyosmani PTAL.

addyosmani:
@robdodson could you rebase and will take another look? Thanks for reinvestigating!

robdodson:
Rebased. Looks like someone already took care of the scroller issue so this just adds Polymer.dom calls to the rest of the file. @addyosmani PTAL

addyosmani:
Verified as working in Firefox, Safari, Chrome. Going to test IE and we should then be good to go :+1:

addyosmani:
Thanks again, @robdodson! :star: :cake:

samccone:
awww yes :+1:

PR #437, opened 2015-10-14
Updated "on-click" to "on-tap"

WesleyDRobinson:
Following the suggestion from the Polymer docs.

>"...tap should be used instead of click for the most reliable cross-platform results."
>https://www.polymer-project.org/1.0/docs/devguide/events.html#gestures


samccone:
LGTM just need a manual test

addyosmani:
Locally verified that `on-tap` appears to be working fine for PSK in desktop browsers. We might want to do some additional testing on actual devices before we land this.

PR #444, opened 2015-10-16
fix(README.md): change the attribute name of the cache file

felixzapata:
Hi, I've updated the attribute name in the section about how to enable service worker support. 

The new version of `platinum-sw-register` use a different name. Now is `cache-config-file` but in the current README file says `precache-file` .

Regards.


addyosmani:
Verified. Thanks for the PR!

PR #458, opened 2015-10-23
Refactor task cache-config

felixzapata:
If you use the package `glob-all` is not necessary to make a push to the array with the files. This way, you can add an array of patterns to the glob function.

Regards.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "connect-history-api-fallback": "^1.1.0",
     "del": "^1.1.1",
     "glob": "^5.0.6",
+    "glob-all": "^3.0.1",
     "gulp": "^3.8.5",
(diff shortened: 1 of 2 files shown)


samccone:
this is great good stuff @felixzapata

samccone:
Landed in 
77b4ab90c179c33715d495877e44f1f9332f3077

thanks again :sparkling_heart:

package.json
```diff
@@ -5,6 +5,7 @@
     "connect-history-api-fallback": "^1.1.0",
     "del": "^1.1.1",
     "glob": "^5.0.6",
```
robdodson: it doesn't look like `glob` is being used by gulp any longer. Can we remove this @samccone ?

felixzapata: yes,  `glob` can be remove

PR #465, opened 2015-10-27
css bloat?

gertcuykens:
```
  .bg {
      background: white;
    }
```


gertcuykens:
And #overlay also i think

addyosmani:
Locally verified that both of these changes should be safe to make. LGTM. Thanks!

Issue #422, opened 2015-10-08
pngs in App/images, not moving into Dist/images


gedw99:
**Environment**
node v0.12.7
polymer tag 1.1.1
OSX

**Die problemo !**
Its very simple, but gulp is just not copying the pngs (chrome-touch-icon-192x192.png for eg) from app/images --> dist/images

I tried a gulp clean. No luck

i just some more checks. Its very flackey, so images get copied and not others.
I think there must be some caching going on somewhere. Gulp being too smart ??

Any ideas ??

addyosmani:
Interesting. Could you try reinstalling your dependencies `npm cache clean && npm install` and then run `gulp` again to see if that improves the copying situation?

I just pulled PSK master and it appears to be copying the images to dist correctly (OSX here too):

<img width="442" alt="screen shot 2015-10-08 at 19 10 52" src="https://cloud.githubusercontent.com/assets/110953/10375453/8a458c54-6df0-11e5-81e3-c9c939140599.png">

gedw99:
@addyosmani 

I just tried a new polymer starter kit and it works perfectly.
But thats not what the problem IS.
I have changes the icons in the apps/images dir,
then i delete the dist folder (to be sure)
then i run "gulp"
then in the dist/images, the images are the original ones, NOT the ones in the app/images.

Hence why i mentioned there seems to be caching going on.

Dragod:
I had same problem with the chrome-touch-icon-192x192.png

gedw99:
I just tried the npm clean and reinstall of npm and bower components

Did not work.
The dist/images has the orginal ones, rather than the ones from app/images

cant work it out

samccone:
@gedw99 Do you get any error messages when you run npm i?

gedw99:
@samccone 

None.

gedw99-MacBook-Pro:static apple$ pwd
/Users/apple/Workspace/go/src/bitbucket.org/gedw99/berlininfusions-web/static
gedw99-MacBook-Pro:static apple$ npm i
gedw99-MacBook-Pro:static apple$

gedw99:
I cross checked that i am not going mad

made a script to copy the images form APP --> DIST, and i then see the real images

not sure whats up.

I also am using the latest gulp script from the latest polymer release:
https://github.com/PolymerElements/polymer-starter-kit/releases/tag/v1.1.1

addyosmani:
[...]

PR #471, opened 2015-10-29
Disable caching for imagemin task (#422)

addyosmani:
This is more of a stop-gap than anything, but should alleviate some of the issues we saw over in #422. r: @chuckh 

I haven't been able to understand why SVG files aren't being copied over. We're following the same pattern used in Web Starter Kit and the gulp-imagemin docs.


chuckh:
LGTM. It seems svgs are no longer an issue.

PR #482, opened 2015-11-04
Remove extraneous `span` tags

hph:
As of Polymer 1.2 these span tags are not necessary anymore and we should provide a good example for users of this starter kit.


hph:
Signed!

robdodson:
LGTM

PR #484, opened 2015-11-05
Add comment to .gitattributes

kaycebasques:
Just the start of a possibly bigger drive to document the application well :)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
+# convert line endings to LF (a.k.a. "normalize")
 * text=auto


samccone:
Hey @kaycebasques I think .gitconfig files are OK not having comments, since the info is a single search away (plus I think anyone looking at this file would already sorta be on the path of learning)

thoughts?

kaycebasques:
I'm cool with that. You can close this. My thought was that I would comment
stuff as I encounter it. But I think I'm getting the gist of your
philosophy: only comment the stuff unique to PSK that the users won't be
able to find elsewhere in order to keep readability high.

samccone:
:+1:

PR #486, opened 2015-11-07
Use gulp-minify-inline to minimize the elements.vulcanized.html after generation

moderndeveloperllc:
As the `vulcanize` process inlines the JS and CSS for elements, the resulting `elements.vulcanized.html` contains several hundred KB of empty space. By using gulp-minify-inline on the vulcanized file, the size goes from 447K down to around 267K for the default install. Even with gzipped calls, this can be quite a bit of pageload reduced.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -171,6 +171,7 @@ gulp.task('vulcanize', function () {
       inlineCss: true,
       inlineScripts: true
     }))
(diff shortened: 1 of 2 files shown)


samccone:
aww this is sooo goood A+ work @moderndeveloperllc :dancer:

PR #494, opened 2015-11-09
[Docs] Update ES2015 docs with latest in gulpfile.js changes

moderndeveloperllc:
Changing to jscs means change jshint to `lint` for function calls; use `dist()` instead of `dist`


samccone:
great work here @moderndeveloperllc 

landed this in 7bdbe7bbdc2e83d9c75140411003ec1b65178b85

addyosmani:
Awesome work!

PR #501, opened 2015-11-11
Babel now requires the ES2015 preset.

birkir:
Checkout the latest docs for gulp-babel: https://github.com/babel/gulp-babel


samccone:
great catch @birkir :)

samccone:
I blame @sebmck for this! :fire:

PR #502, opened 2015-11-11
Typo fix - example missing quotes

wejrowski:
Just fixing what appears to be a typo as I was playing with the example code.


wejrowski:
I signed it!

samccone:
yessssir, great spot!

Issue #469, opened 2015-10-29
Don't include __MACOSX or .DS_Store in release


fredj:
A `__MACOSX/` directory and `app/.DS_Store` file are present (at least) in `polymer-starter-kit-light-1.1.1.zip`

fredj:
Only `polymer-starter-kit-light` is affected, not `polymer-starter-kit`

addyosmani:
This is my fault. It's an artefact of our light release process. For future releases, I'll make a note to ensure our tooling needs to avoid those two files.

fredj:
See also #464

robdodson:
Btw I added a PR to address the .DS_Store issue https://github.com/PolymerElements/polymer-starter-kit/pull/618

fredj:
Thanks !

PR #519, opened 2015-11-18
Add .DS_Store to .gitignore for OSX user

timwangdev:


timwangdev:
I signed it!

abdonrd:
I think that the `.DS_Store` should be in the `.gitignore_global`.

timwangdev:
See also #469

samccone:
+1 for adding to your personal global ignore

will leave the final call up to @addyosmani

timwangdev:
@abdonrd @samccone I admit that adding `.DS_Store` to `.gitignore_global` would be more reasonable. I'll close this one.

samccone:
thanks for the idea though @timwangdev :)

PR #525, opened 2015-11-20
Fix 300ms click delay for menu

robdodson:


samccone:
wait... what I am so confused about this.. I thought on-tap was the right way to do things

robdodson:
@samccone not in this case. We have handlers listening for `on-tap` to know when to close the drawer. So it hears the touch events, closes the drawer, then 300ms on iOS the `click` event fires, but now there's no longer an anchor in that position (because the drawer closed), so the link doesn't get clicked and we don't change pages. Discussed this with the Polymer team and the above fix is what they recommended.

samccone:
so, having a blanket on-tap listener on the drawer regardless of the target seems like a bug in the drawer code to me.. like yes this is a fix, however it feels more like a hack around event propagation

robdodson:
@samccone the `on-tap` listener was on the anchor clicks (not the drawer) but that's been switched to `on-click`

samccone:
ok :)

I wonder if we might just want to handle the "click" ourselves in
https://github.com/PolymerElements/polymer-starter-kit/blob/7082366bc258ae748837d3adfe754e70955180f9/app/scripts/app.js#L73-L78

and then 

```js
e.preventDefault()
page.visit(e.currentTarget.....)
```

robdodson:
@samccone PTAL. I moved the work into routing and also rewrote the routes so we're not duplicating the middleware usage.

app/scripts/app.js
```diff
@@ -69,17 +69,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
...
-    }
-  };
-
   // Scroll page to top and expand header
   app.scrollPageToTop = function() {
     app.$.headerPanelMain.scrollToTop(true);
   };
 
+  app.closeDrawer = function() {
+    app.$.paperDrawerPanel.closeDrawer();
```
chuckh: @robdodson the `app.$.paperDrawerPanel.closeDrawer()` will always be called if narrow or not, but won't close the drawer when not narrow?

robdodson: `closeDrawer` doesn't seem to be doing anything fancy (https://github.com/PolymerElements/paper-drawer-panel/blob/master/paper-drawer-panel.html#L543) so I was thinking of just leaving out the narrow check to slim down the code. ¯\_(ツ)_/¯

chuckh: SGTM

addyosmani: SGTM2

addyosmani:
LGTM but assigning to @samccone for a final call on the don't merge yet before we land this. Nice work, @robdodson :star:

samccone:
:+1: good work all around :+1:

Issue #532, opened 2015-11-24
gulp deploy gives error


VikramTiwari:
```
Task 'deploy' is not in your gulpfile
```

samccone:
@VikramTiwari did you see this gulp task documented anywhere? I am curious as to what made you think this would work :)

VikramTiwari:
haha well, my bad. Should have gone through documentation first, which lists the correct method.

Though `gulp deploy` is the obvious command that I would use at times, hence the mistake. Close it if you want to :P

PR #533, opened 2015-11-24
Update gulpfile.js

VikramTiwari:
for solution to https://github.com/PolymerElements/polymer-starter-kit/issues/532


samccone:
https://github.com/PolymerElements/polymer-starter-kit/issues/532#issuecomment-159179379

PR #537, opened 2015-11-24
Add travis-runner-sh to .travis.yml

chuckh:
This will cause travis-runner-sh to auto run.  @samccone please review :smile:


samccone:
heh this is important I was just scratching my head as to why it was not working :dancer:

PR #539, opened 2015-11-24
Fix sed syntax so it will work with both Mac & Linux

chuckh:
See http://stackoverflow.com/a/22084103/2403875. Tested on Mac and it worked!  Should now work for travis-ci it uses Ubuntu.


samccone:
ha what a gem :gem:

PR #545, opened 2015-11-28
Use inline_script mode when minifying vulcanized HTML

TimothyGu:
Ideally this should be fixed in gulp-minify-inline, but for now this should fix some issues related to firebase elements as reported by @johnwhitton on the Polymer Slack.

To go into more detail, before this PR UglifyJS automatically transforms [`'";\x3c/script>'`][1] to `'";</script>'`, which breaks because the string is in an inline `<script>` tag. After this PR, the same string gets transformed into `'";<\/script>'`.

Note that this only applies to strings; UglifyJS does not provide any protection against more sophisticated `</script>` escaping like that in Polymer/vulcanize#280.

[1]: https://github.com/firebase/firebase-bower/blob/master/firebase.js#L171

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -193,7 +193,15 @@ gulp.task('vulcanize', function() {
       inlineCss: true,
       inlineScripts: true
     }))
-    .pipe($.minifyInline())
+    .pipe($.minifyInline({
+      js: {
+        output: {
+          // jshint ignore:start
+          inline_script: true // jscs:ignore requireCamelCaseOrUpperCaseIdentifiers
+          // jshint ignore:end
+        }
+      }
+    }))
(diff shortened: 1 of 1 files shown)


samccone:
@TimothyGu mind rebasing this real quick for us :)

thanks!

moderndeveloperllc:
I think @ebidel just removed this call. Double check with latest master

samccone:
Oh yeah heh we did just revert this bit goibg to close for now but we are looking into better options around this problem area

TimothyGu:
Or we could just disable minifying the CSS. And it would probably be better if we used gulp-htmlmin which allows minifying HTML in addition to inline CSS (which we can disable) and JavaScript.

PR #553, opened 2015-12-02
Update href paths in index.html

moderndeveloperllc:
If you don't add the `{{baseUrl}}`, the href paths are wrong after you go to a specific user's page - i.e. when on `http://localhost:5000/users/Addy`, the **Users** link will take you to `http://localhost:5000/users/users`

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -79,12 +79,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             <span>Home</span>
           </a>
 
-          <a data-route="users" href="users">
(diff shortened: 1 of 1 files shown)


samccone:
heh, this is a good catch @moderndeveloperllc 

just want to take this for a spin before merge

moderndeveloperllc:
@samccone I'll admit I haven't tested with anything other than default links with a baseUrl of `/`. *Should* work though.

samccone:
working for me :clap: thanks @moderndeveloperllc

PR #554, opened 2015-12-02
Fix cache-config-file in README

JeremybellEU:
If you would copy-paste the HTML code in the README.md, it would generate a 404 as `precache.json` was not found.
In `index.html`, this is `cache-config.json` which does not 404.


addyosmani:
Lgtm. Thanks!

Issue #552, opened 2015-12-02
Add travis recipe checker


chuckh:
Add to travis-runner.sh a checker for our recipes. :computer: :mag: :white_check_mark: 

from: @samccone on Slack
> soooooo if we format the code changes as patches :heavy_plus_sign: :heavy_minus_sign: 
and on CI walk all the examples and try and apply the patches...
and they do not cleanly go in... we blow up :boom:  
it will force us to update the recipes :scroll:  as we make code changes
so the burden of maintenance is moved into our normal workflow... it will also allow easier maintenance of recipes over time :clock1: :clock2: :clock4:

addyosmani:
:+1:

chuckh:
One possible approach is to use gemini utility for regression testing of web pages using screenshots https://github.com/gemini-testing/gemini.  Vaadin project is planning to use it. Per this [comment](https://plus.google.com/u/0/114229924035668909209/posts/evKaqv8mc3Q?cfem=1).

chuckh:
Closing in favor of PSK2 which does not have recipes.  Instead of doing recipes we can create custom templates.

PR #555, opened 2015-12-02
Add friendly error message for missing dotfiles

### LINKED ISSUE #544: TypeError: Cannot convert undefined or null to object in task lint
Download 1.2 directly, then did npm install and bower install, then do "gulp serve", and here is the error output:
```
[12:37:13] Starting 'lint'...
[12:37:14] 'lint' errored after 1.18 s
[12:37:14] TypeError: Cannot convert undefined or null to object
    at Function.keys (native)
    at copyConfiguration (/Users/cnbuff410/node_modules/gulp-jscs/node_modules/jscs/lib/config/configuration.js:1181:12)
    at NodeConfiguration.Configuration._processConfig (/Users/cnbuff410/node_modules/gulp-jscs/node_modules/jscs/lib/config/configuration.js:474:5)
    at NodeConfiguration.Configuration.load (/Users/cnbuff410/node_modules/gulp-jscs/node_modules/jscs/lib/config/configuration.js:238:10)
    at StringChecker.configure (/Users/cnbuff410/node_modules/gulp-jscs/node_modules/jscs/lib/string-checker.js:67:29)
    at Checker.configure (/Users/cnbuff410/node_modules/gulp-jscs/node_modules/jscs/lib/checker.js:28:39)
    at Object.module.exports (/Users/cnbuff410/node_modules/gulp-jscs/index.js:35:10)
    at Gulp.<anonymous> (/Users/cnbuff410/gulpfile.js:124:11)
    at module.exports (/Users/cnbuff410/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:34:7)
    at Gulp.Orchestrator._runTask (/Users/cnbuff410/node_modules/gulp/node_modules/orchestrator/index.js:273:3)
```
It's caused by this line in lint task:

  .pipe($.jscs())

ohanhi:
Fixes #544.

This provides much more human-readable error messages for when certain dotfiles (namely `.jscsrc` or `.jshintrc`) are missing from the file system.

With the improvement:
screen shot 2015-12-02 at 23 26 38

Previously:
screen shot 2015-12-02 at 23 25 47

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -114,7 +114,8 @@ gulp.task('lintfilesExist', function(cb) {
 
   var checked = 0;
   necessaryFiles.forEach(function(file) {
-    fs.stat(__dirname + '/' + file, function(err, stats) {
+    var filePath = path.join(__dirname, file);
+    fs.stat(filePath, function(err, stats) {
       if (err || !stats.isFile()) {
         filesMissing.push(file);
       }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -105,8 +105,33 @@ gulp.task('elements', function() {
   return styleTask('elements', ['**/*.css']);
 });
 
+// Check the files necessary for linting exist.
+// Without this, the 'lint' task can fail with a very non-descriptive message:
+// "TypeError: Cannot convert undefined or null to object"
+gulp.task('lintfilesExist', function(cb) {
+  var necessaryFiles = ['.jscsrc', '.jshintrc'];
+  var filesMissing = [];
+
+  var checked = 0;
+  necessaryFiles.forEach(function(file) {
+    fs.stat(__dirname + '/' + file, function(err, stats) {
+      if (err || !stats.isFile()) {
+        filesMissing.push(file);
+      }
+      checked++;
+      if (checked === necessaryFiles.length) {
+        if (filesMissing.length === 0) {
+          cb();
+        } else {
+          cb('Cannot lint! These necessary files are missing: ' + filesMissing.join(', ') + '.\n');
+        }
+      }
+    });
+  });
+});
+
 // Lint JavaScript
-gulp.task('lint', function() {
+gulp.task('lint', ['lintfilesExist'], function() {
   return gulp.src([
       'app/scripts/**/*.js',
       'app/elements/**/*.js',


ohanhi:
I signed it!

gulpfile.js
```diff
@@ -105,8 +105,33 @@ gulp.task('elements', function() {
...
+// Check the files necessary for linting exist.
+// Without this, the 'lint' task can fail with a very non-descriptive message:
+// "TypeError: Cannot convert undefined or null to object"
+gulp.task('lintfilesExist', function(cb) {
+  var necessaryFiles = ['.jscsrc', '.jshintrc'];
+  var filesMissing = [];
+
+  var checked = 0;
+  necessaryFiles.forEach(function(file) {
+    fs.stat(__dirname + '/' + file, function(err, stats) {
```
samccone: path.join() here :)

ohanhi:
@samccone Is this ok?

gulpfile.js
```diff
@@ -105,8 +105,34 @@ gulp.task('elements', function() {
...
 
+// Check the files necessary for linting exist.
+// Without this, the 'lint' task can fail with a very non-descriptive message:
+// "TypeError: Cannot convert undefined or null to object"
+gulp.task('lintfilesExist', function(cb) {
+  var necessaryFiles = ['.jscsrc', '.jshintrc'];
+  var filesMissing = [];
+
+  var checked = 0;
+  necessaryFiles.forEach(function(file) {
```
samccone: can we do 

necessaryFiles.map(function(v){ return path.join(__dirname, v); })

reduce or filter down

then do the error check?


(I think in this case it is totally ok to use statSync just to keep the complexity low)

samccone:
@ohanhi just ping me if you have any questions or whatever, or let me know any thoughts

thanks again for this great PR!

robdodson:
I'd like to offer a dissenting opinion here. The gulpfile for this project is already extremely complicated and due for some house cleaning. I would vote against adding another task to check for missing files. We have a record of the problem and the fix documented in #544.

samccone:
The flip side is that this makes an error, which several people now have hit, into a human readable easy to debug locally issue as compared to filing an issue on the bug tracker.

For that reason I am :+1: to it. Most people never touch the gulp file, and those that do are applying recipes.. which cause issues for other reasons .. but we have an idea how to fix that set of issues https://github.com/PolymerElements/polymer-starter-kit/issues/552

so, +1 to better errors --- the complexity argument is valid however I think the only reason why I see it as an issue are for the reasons state in 552

robdodson:
I think this can be improved with a few lines of instructions in the README vs adding code to the project to support it.

gulpfile.js
```diff
@@ -105,8 +105,34 @@ gulp.task('elements', function() {
   return styleTask('elements', ['**/*.css']);
 });
 
+// Check the files necessary for linting exist.
+// Without this, the 'lint' task can fail with a very non-descriptive message:
+// "TypeError: Cannot convert undefined or null to object"
+gulp.task('lintfilesExist', function(cb) {
```
robdodson: Instead of creating a separate task for this, could we move the code inside the lint task itself? @samccone wdyt?

robdodson: I did a variation on this which seems to work ok: https://gist.github.com/robdodson/9219e1b5042eeca14923

I'm not an expert in throwing errors in Gulp but here's what the output looks like:

screen shot 2015-12-06 at 9 27 18 am

samccone: personally I think a separate task is pretty nice, easier to test... debug ... ect ect

samccone:
moved to #580

PR #558, opened 2015-12-03
Fix thumbnail link in mobile-chrome-apps.md

floogulinc:
The thumbnail now goes to the video when clicked.


samccone:
great catch and fix @floogulinc :) thanks a ton for the PR!!!

Issue #561, opened 2015-12-04
Polymer Starter Kit Demo Does Not Work in Chrome


paranoid-android:
Hello,

The [Polymer Starter Kit Demo](http://polymerelements.github.io/polymer-starter-kit/) does not render properly in Chrome. The menu items are not spaced properly. In addition, the refresh and search icons are left-aligned instead of right-aligned.

I am experiencing this issue in: 

Chrome v: 47.0.2526.73 (64-bit)
Mac OS X El Capitan

chuckh:
Looks like  `iron-flex-layout` version 1.2.2 is causing the problem. Reverting back to 1.2.1 seems to fix it.  I am working on temporary fix reverting back to version 1.2.1 now.

chuckh:
You can see this issue on master branch by doing a `gulp serve:dist`, it does not happen with `gulp serve`.

Also, I manually deployed to  [Polymer Starter Kit Demo](http://polymerelements.github.io/polymer-starter-kit/) this fix. It fixes menu issues for Chrome, Firefox and Safari, but not IE11 and Edge issue #550.

addyosmani:
Sounds like we need to tag a patch release to address this now that the above commit is in master.

opus1269:
I have the same issue in another app. The problem is with call to minifyInline. Something in the refactoring of iron-flex-layout CSS selectors is causing issues. You can use the css: false option in minifyInline to fix it for now.

```js
// vulcanize for production
gulp.task('vulcanize', function() {
	return gulp.src(path.elements.src + 'elements.html')
	.pipe(plugins.vulcanize({stripComments: true, inlineCss: true, inlineScripts: true}))
	.pipe(plugins.crisper({scriptInHead: false}))
	.pipe(plugins.if('*.html', plugins.minifyInline({css: false}))) // can't handle polymer css
	.pipe(plugins.if('*.js', plugins.uglify()))
	.pipe(gulp.dest(path.elements.dist));
});

robdodson:
We removed minifyInline in the most recent release to fix this

opus1269:
I found the root cause of this and submited an issue on iron-flex-layout:
https://github.com/PolymerElements/iron-flex-layout/issues/60

addyosmani:
This issue should be (for now) addressed in the 1.2.1 release available from the Releases page.

robdodson:
@addyosmani are you thinking we should close it?

samccone:
[...]

Issue #559, opened 2015-12-03
Distorted style when doing gulp serve:dist


vishnudevk:
When i am doing **gulp serve:dist** the page looks different from **gulp serve**.
components are distorted in dist. same issue is coming while deploying in firebase also
Please find attached screenshots
with  **gulp serve**.
![serve](https://cloud.githubusercontent.com/assets/4205634/11559513/8f1e34d6-99df-11e5-9521-131a51300b5a.PNG)
with  **gulp serve:dist**
![dist](https://cloud.githubusercontent.com/assets/4205634/11559515/93142a0a-99df-11e5-95c0-8b04d9067c42.PNG)

addyosmani:
Fixed by https://github.com/PolymerElements/polymer-starter-kit/pull/563. Tag coming Monday.

Issue #557, opened 2015-12-02
Build process breaks layouts in 1.2.0


vaadin-marcus:
Layouts break when running app with `gulp serve:dist`, see screenshots below:

Running with `gulp serve`: 
<img width="954" alt="serve" src="https://cloud.githubusercontent.com/assets/883698/11547125/81bfe944-9906-11e5-90ed-901e49c4fac1.png">

Running with `gulp serve:dist`
<img width="945" alt="dist-serve" src="https://cloud.githubusercontent.com/assets/883698/11547127/83fbf2f2-9906-11e5-96a0-4c3459a05542.png">

Steps to reproduce:
1. Download PSK 1.2.0
2. npm install && bower install && gulp serve:dist
3. Layouts are broken

Possibly/probably related to https://github.com/Polymer/polymer/issues/3126

chuckh:
I see this problem now also. You can see the problem live at http://polymerelements.github.io/polymer-starter-kit/.

This does not happen on https://polymer-starter-kit.firebaseapp.com/ which was built a couple days ago using 1.2.0.  Might be an upstream issue related to deprecating /deep/ work going on now. See https://blog.polymer-project.org/announcements/2015/12/01/deprecating-deep/.

addyosmani:
Fixed by https://github.com/PolymerElements/polymer-starter-kit/pull/563. We'll tag a new release Monday.

Issue #550, opened 2015-12-02
Internet Explorer style issues in PSK 1.2.0


ghost:
There are some heavy style issues in internet explorer for the new PSK. (See attachment).
![untitled](https://cloud.githubusercontent.com/assets/13914108/11527782/2d1d1e3e-98e3-11e5-9395-dfb0f11e4656.png)

addyosmani:
Which version of IE are you testing against?

BorntraegerMarc:
Newest version of internet explorer 11 and experience the same issue on edge.

chuckh:
I got the same results on Windows 10 using IE 11 and Edge. Not sure if it is a PSK issue or up stream issue.

BootsSiR:
I started to make an application with PSK about four days ago and just out of curiosity, I opened it up in IE 11/Edge and it's badly broken.  The first two menu items in the drawer don't show up on load, however if I resize the window, they pop into view.  Also navigation doesn't seem to work.  I don't get any javascript errors in the console but it seems badly broken.

BorntraegerMarc:
@addyosmani Should I open another issue for IE/Edge support in general? And if yes, where should i open it?

addyosmani:
We can use this issue for it. Atm, we'll need to assess if this is an issue with PSK itself or whether this was caused by one of the `paper-*` dependencies we use. 

We'll want to test:

https://elements.polymer-project.org/elements/paper-scroll-header-panel?view=demo:demo/transform-header-5.html&active=paper-scroll-header-panel

https://elements.polymer-project.org/elements/paper-drawer-panel?view=demo:demo/index.html&active=paper-drawer-panel

https://elements.polymer-project.org/elements/paper-menu?view=demo:demo/index.html&active=paper-menu

for breakage

BorntraegerMarc:
So I've tested the pages and attached the screenshots (always IE first)... It looks like edge can handle every good. The layout seems right and it behaves as it should... Although IE can't even display the element.
![image](https://cloud.githubusercontent.com/assets/13914108/11560523/6c9449fa-99c1-11e5-89d8-574d67bf98ea.png)

![image](https://cloud.githubusercontent.com/assets/13914108/11560529/71731096-99c1-11e5-96d0-3aa04815a9c8.png)

![image](https://cloud.githubusercontent.com/assets/13914108/11560535/76c5b8aa-99c1-11e5-9722-dd73a762b3fa.png)

[...]

PR #563, opened 2015-12-04
Fix menu layout issues

chuckh:
Fixes issues #561, #559, #557 by reverting back to iron-flex-layout
version 1.2.1, version 1.2.2 causes this problem. Still need to find long term fix.

Fixes layout issues for Chrome, Firefox and Safari, but not IE 11 or
Edge (#550).

r: @addyosmani / @samccone

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,8 @@
     "paper-elements": "PolymerElements/paper-elements#^1.0.1",
     "platinum-elements": "PolymerElements/platinum-elements#^1.1.0",
     "polymer": "Polymer/polymer#^1.2.0",
-    "paper-menu": "PolymerElements/paper-menu#4fecb43601"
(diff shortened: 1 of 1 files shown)


chuckh:
I already manually deployed to Github Pages to fix demo. You can see it working at polymerelements.github.io/polymer-starter-kit.

addyosmani:
Lgtm

robdodson:
Change is outlined in this blog post:
https://blog.polymer-project.org/announcements/2015/12/01/deprecating-deep/

More than likely we're relying on some element to import the flex layout classes and now that element is no longer doing so. We should fix PSK instead of going back a version

samccone:
@addyosmani we should cut a patch release, so that people can a valid base to go from.

I agree @robdodson :) Mind opening up a new issue detailing the work so that anyone can hop on it?

PR #568, opened 2015-12-05
Move bower_components. Make vulcanize easier

robdodson:
Fixes #562
Should help #512

- Move bower_components into app dir
  - Because we mount the bower_components directory it makes vulcanizing totally insane (hence all the weird stuff where we copied over all of bower_components and vulcanized in the dist directory, instead of vulcanizing in app and writing the output to dist. Doing it this way also screws up tools like `web-component-shards` and Superstatic (as mentioned in #512). I realize it's [a Yeoman convention](http://yeoman.io/blog/bower_components-in-project-root.html) to have bower_components outside of app, but for Polymer projects that makes life very difficult.
- Remove code to rename `elements.html` to `elements.vulcanized.html`. Instead, `dist/elements/elements.html` is just the vulcanized version of `app/elements/elements.html`
- Tell `optimizeHtmlTask` to ignore `bower_components`
- Tell `platinum-sw-register` where to find bootstrap dir with `base-uri` attribute

PR #569, opened 2015-12-06
Add dotfile warning to readme

robdodson:
Proposing this as an alternative to https://github.com/PolymerElements/polymer-starter-kit/pull/555
Hopefully improves https://github.com/PolymerElements/polymer-starter-kit/issues/544

Addy has added a write up to the release notes, and hopefully adding this warning to the README also makes it very clear that the user should be on the lookout for those dotfiles. I would much prefer we try to go this route and see if folks still have problems vs. adding new code to the gulpfile to check for missing stuff.

`<tangent>`
I've been feeling that the gulpfile is getting really large and difficult for users to modify and, equally important, for us to safely maintain. https://github.com/PolymerElements/polymer-starter-kit/pull/568 is an attempt to correct some of this. I'd love if we could do some aggressive pruning to get it down as small as possible so it isn't so intimidating.
`</tangent>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ To take advantage of Polymer Starter Kit you need to:
 
 **Intermediate - Advanced**: Use the full version of Polymer Starter Kit. This comes with all the build tools you'll need for testing and productionising your app so it's nice and lean. You'll need to run a few extra commands to install the tools we recommend but it's worth it to make sure your final app is super optimised.
 
+:warning: **Important**: The intermediate/advanced version contains dotfiles (files starting with a `.`). If you're copying the contents of the Starter Kit to a new location make sure you bring along these dotfiles as well! On Mac, [enable showing hidden files](http://ianlunn.co.uk/articles/quickly-showhide-hidden-files-mac-os-x-mavericks/), then try extracting/copying Polymer Starter Kit again. This time the dotfiles needed should be visible so you can copy them over without issues.
+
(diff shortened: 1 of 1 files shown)


addyosmani:
Commenting on the tangent: I agree. This is actually an opinion I hold with both Polymer Starter Kit and Web Starter Kit. The Gulpfiles are almost unwieldy. One thing we may explore next quarter is having a split Gulpfile tasks repo that both projects can pick and choose from. This may reduce complexity and overlap whilst making it easier to reason what is going on.

chuckh:
Change to readme LGTM. :+1: 

I agree with @robdodson and @addyosmani  on <tangent>.  The gulp file is complex, pruning and splitting tasks out in separate files could make it simpler to understand and maintain. Having a shared repo for both PSK and WSK  makes sense.

Polymer Starter Kit Plus has split out gulp tasks. See https://github.com/StartPolymer/polymer-starter-kit-plus/tree/master/tasks.

addyosmani:
I'm :+1: on readme change too (thanks for reviewing, @chuckh!). @samccone do you feel that a README note would be sufficient here or that there's enough chance that folks aren't going to read it that having a build-time warning is better?

samccone:
I am for this change, however I am also for the gulpfile change, I think they are both very valuable to two different sets of people

chuckh:
> I am for this change, however I am also for the gulpfile change, I think they are both very valuable to two different sets of people

I agree. Landing this. Can handle error message with pr #555.

PR #572, opened 2015-12-07
Small tweak to Firebase recipe

robdodson:
`hashbang: false` is the default for `page.js` so you can just remove it entirely and call `page()`. @chuckh PTAL


robdodson:
Ignore the CLA bot. I used the GitHub text editor to edit the file which is why it's complaining

samccone:
thx rob

chuckh:
LGTM

PR #586, opened 2015-12-12
Add "en" html lang

abdonrd:


robdodson:
LGTM, all text that comes with the page is currently written in English so seems like a reasonable default.

PR #588, opened 2015-12-12
Indent all html files

abdonrd:


robdodson:
Did you run the files through a tool like sublime beautify?

abdonrd:
No, I indent the `head` and `body` manually.

samccone:
To merge this change I would like to see the following:

1. An automated lint process that will cause failures on CI if the indentation is wrong
2. The equivalent of jshint but for html

----

The reason why this change is not scalable as it stands is that the burden of formatting verification will fall on whoever reviews the next change, and tiny formatting nits are often not caught.

Thanks for the interest and PR here @abdonrd, let me know your thoughts on what I said ^

abdonrd:
@cdata I am strongly in favor of automating this. 👍

Maybe [HTMLHint](https://github.com/yaniswang/HTMLHint)? There are an [Atom package](https://atom.io/packages/linter-htmlhint).

And a style guide for PSK and Polymer elements would be awesome.

cdata:
@abdonrd We will be adding Polylint to our elements' CI flow in the future, but I don't think that it will be picky about indentation issues. @garlicnation can comment on if that is in the cards or not.

When it comes to picking a linter for automation, the most important thing IMO is that its rules are configurable. HTML linters can be particularly picky about some details (e.g., doctype validation) and I don't think we should force ourselves to adhere strictly to everything the linter prefers without some discussion.

abdonrd:
@cdata Personally, I like the idea that in a project, all follow the same style guidelines. Especially when the project has a big community. But I also understand your position.

cdata:
@abdonrd I completely agree. What I mean to say is that the style guidelines should be chosen by the project's contributors, not by the linter.

abdonrd:
@cdata ah, okey! I completely agree too. :grin:

samccone:
going to close this for now, @abdonrd feel free to open an issue with some possible linters for us, or any thoughts around this.

Thanks a bunch for this initial spike of work!

PR #594, opened 2015-12-15
Update gulp clean task

m-nel:
Only clean the contents of the dist dir instead of deleting the dist dir itself.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -232,7 +232,7 @@ gulp.task('cache-config', function(callback) {
 
 // Clean output directory
 gulp.task('clean', function() {
-  return del(['.tmp', dist()]);
+  return del(['.tmp', dist('**'), ['!', dist()].join('')]);
 });
 
 // Watch files for changes & reload


m-nel:
I signed it!

robdodson:
Hey thanks for the PR. Can you explain why you want to keep the directory
in place vs just deleting it?

m-nel:
Hey Rob,

__Gladly:__
I'm in the process of making a docker container for running the builds (and the rest of the gulp tasks).

In order to get the distributable content, the dist folder needs to be mounted on the container.
But since the 'clean' task does an `rm` on the directory it creates a conflict with the mount mentioned above :(

__re: PR__
I don't know if the empty dist folder might be a problem/nuisance for other developers though

robdodson:
Hm so I'm thinking that maybe we shouldn't merge this, only because it adds complexity to the gulpfile and I'm assuming most users are not deploying containers or under the same constraints. For those who are, they can refer to this ticket for guidance. @samccone @addyosmani @chuckh do either of you have an opinion on this?

chuckh:
@robdodson I agree with you.

> I'm thinking that maybe we shouldn't merge this, only because it adds complexity to the gulpfile and I'm assuming most users are not deploying containers or under the same constraints. For those who are, they can refer to this ticket for guidance

robdodson:
OK, going to close for now. Happy to reopen if others disagree. Also, thank you @m-nel for working on the PR

PR #611, opened 2015-12-17
Relax jscs indentation rules

robdodson:
JSCS is barking at a lot of our users, many of whom may be unfamiliar with it. This relaxes JSCS rules around indentation a bit while we sort out the best way to deal with extracted JS

@samccone PTAL


samccone:
@robdodson looks like maybe you are not based off of latest master?

robdodson:
boo

robdodson:
do over

samccone:
:dancers:

Issue #436, opened 2015-10-13
Why there are two custom properties (--text-primary-color and --primary-text-color) with such similar names?


abdonrd:
In the [app/styles/app-theme.html](https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/styles/app-theme.html):

```css
  /* Application theme */

  :root {
    --dark-primary-color: #303F9F;
    --default-primary-color: #3F51B5;
    --light-primary-color: #C5CAE9;
    --text-primary-color: #ffffff; /*text/icons*/
    --accent-color: #FF4081;
    --primary-background-color: #c5cae9;
    --primary-text-color: #212121;
    --secondary-text-color: #727272;
    --disabled-text-color: #bdbdbd;
    --divider-color: #B6B6B6;
```

Why there are two custom properties (`--text-primary-color` and `--primary-text-color`) with such similar names?

addyosmani:
This was originally based on https://github.com/PolymerElements/paper-styles/blob/master/default-theme.html. Looking at the spec from the Material Design palette, I do wonder if it's possible for us to make the naming here more clear. @morethanreal or @notwaldorf would either of you have suggestions on how we could improve the variable names used in default-theme to be less confusing?

<img width="849" alt="screen shot 2015-10-13 at 14 41 51" src="https://cloud.githubusercontent.com/assets/110953/10456297/93810fc0-71b8-11e5-8348-f033efaf4bc6.png">

reiaguilera:
:+1:

addyosmani:
Friendly ping @notwaldorf in case you get a chance to check in on this :) I know you've got your hands full!

abdonrd:
Friendly ping to @addyosmani and @notwaldorf. :)

notwaldorf:
Here's the answer I left in https://github.com/PolymerElements/paper-styles/issues/81:
 
I honestly don't know the origin of this, and unfortunately we're now in a situation where both are probably in use, so none can be deleted without being seen as a giant breaking change. :(

`primary-text-color` looks like the right one (87% of #000), and `text-primary-color` is white and I don't know what it's intentions are. I will add a comment to try to discourage people from using it. 

I know this isn't a very satisfying answer, but it's what I got!

addyosmani:
[...]

PR #612, opened 2015-12-18
Update to the new theme custom properties

abdonrd:
Update to the new theme custom properties.
Reference: https://github.com/PolymerElements/polymer-starter-kit/issues/436#issuecomment-163551192

### DIFF
--- a/app/styles/app-theme.html
+++ b/app/styles/app-theme.html
@@ -24,16 +24,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   /* Application theme */
 
   :root {
-    --dark-primary-color: #303F9F;
-    --default-primary-color: #3F51B5;
-    --light-primary-color: #C5CAE9;
-    --text-primary-color: #ffffff; /*text/icons*/
-    --accent-color: #FF4081;
-    --primary-background-color: #c5cae9;
     --primary-text-color: #212121;
-    --secondary-text-color: #727272;
-    --disabled-text-color: #bdbdbd;
-    --divider-color: #B6B6B6;
+    --primary-background-color: #ffffff;
+    --secondary-text-color: #737373;
+    --disabled-text-color: #9b9b9b;
+    --divider-color: #dbdbdb;
+    --primary-color: #3f51b5;
+    --light-primary-color: #c5cae9;
+    --dark-primary-color: #303f9f;
+    --accent-color: #ff4081;
+    --light-accent-color: #ff80ab;
+    --dark-accent-color: #f50057;
 
     /* Components */
 
@@ -74,7 +75,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   }
 
   .paper-menu > .iron-selected {
-    color: var(--default-primary-color);
+    color: var(--primary-color);
   }
 
   paper-menu a {


app/styles/app-theme.html
```diff
@@ -24,16 +24,17 @@
   /* Application theme */
 
   :root {
-    --dark-primary-color: #303F9F;
-    --default-primary-color: #3F51B5;
```
robdodson: `--default-primary-color` is used here: https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/styles/app-theme.html#L77

I just sent in a PR to update that spot with a mixin (which is also using `--default-primary-color`). 

I think this PR should update the above mentioned line to use `--primary-color` and then I can rebase and update my PR to also use `--primary-color`. Does that sound ok?

robdodson:

robdodson:
I only have the one comment above, @abdonrd can you update the PR to address it?

It looks like some of the colors vary ever so slightly from their original values but the page looks pretty similar to me. Here's an updated screenshot:

screen shot 2015-12-19 at 10 13 33 am

It may be so subtle that most folks won't notice.

app/styles/app-theme.html
abdonrd: Ups! :sweat_smile:
Yeah! Sounds good to me!

abdonrd:
Yes, @robdodson, the minor changes are:

`--primary-background-color` from `#c5cae9` to `#ffffff`.
`--secondary-text-color` from `#727272` to `#737373`.
`--disabled-text-color` from `#bdbdbd` to `#9b9b9b`.
`--divider-color` from `#B6B6B6` to `#dbdbdb`.

Now we have the same as Material Design spec. :smile:

abdonrd:
By the way, PR updated! :dizzy:

robdodson:
Change LGTM

robdodson:
I'll wait for one of the other team members to :+1: then we can merge

chuckh:
:+1:

robdodson:
Thanks @abdonrd!

abdonrd:
@robdodson, any news about this?

robdodson:
Pining @notwaldorf for a review

notwaldorf:
`--default-primary-color` is still used in some paper-elements because i haven't got around to go and fix any of them. so yes, if you change `var(--primary-color)` to teal and expect `paper-checkbox` to turn teal, you'll have a bad time.

The `paper-styles` change wasn't a breaking change, because `--default-primary-color` is still backward compatible, but I think this PR puts PSK ahead of all the paper-elements, which still depend on the `1.0.x` bits of `paper-styles` :(

notwaldorf:
I thiiiiiiiink this is fine? Here's what I'm thinking
- you updated `paper-checkbox` 2 months ago. its dependency, `paper-styles/default-theme.html`, is from 2 months ago, so it’s `--default-primary-color` is old and hardcoded
- you update all your dependencies. there’s been no change to `paper-checkbox`, but when you update dependencies, `paper-styles/default-theme` gets updated to the future, because it has a new release (and in their `bower.json`, elements depend on `paper-styles#^1.0.0). in this future, `--default-primary-color` is aliased to `var(-primary-color)`
- so if you pull all the dependencies correctly, you can just update `var(--primary-color)` because all the elements will have the updated `default-theme.html`

ideally all the elements _should_ stop using `--default-primary-color` but i haven't got around to merging this.

robdodson:
Thank you @notwaldorf :+1: :cake: :fireworks:

abdonrd:
Nice! Thanks @notwaldorf!

PR #613, opened 2015-12-18
Rev html-extract and add strip option

robdodson:
I can't recall if the caret will pick up a minor rev like this so i upticked the version just in case

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -2,7 +2,6 @@
   "preset": "google",
   "disallowSpacesInAnonymousFunctionExpression": null,
   "disallowTrailingWhitespace": null,
-  "validateIndentation": null,
   "maximumLineLength": 100,
   "excludeFiles": ["node_modules/**"]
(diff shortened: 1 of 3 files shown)


samccone:
(can we turn the rule back on :))

robdodson:
I guess sooooooo

I turned it back on and rebased. I think I need to update the ES2015 recipe because it fails for other jscs reasons (spaces around the function parens i think). I'll do a separate PR for that

robdodson:
oh hold on let me fix the whitespace rule as well

robdodson:
nvm, I was thinking of the EOL rule. The whitespace rule was already in there. I think we're good to merge.

samccone:
boom

robdodson:
actually the es2015 recipe looks like it's passing. You just have to be careful how you delete the `+` signs so you don't end up with the wrong indentation.

PR #584, opened 2015-12-11
Add auto deploy to Firebase

chuckh:
This adds the auto deploy part. Does not address testing.

Getting this working maybe iterative process like auto deploying GitHub
Pages was.

Addresses auto deploy part of #583 

r: @robdodson @samccone

Issue #512, opened 2015-11-15
Make it easier to work with Firebase Hosting


mbleigh:
Firebase engineer here :fire: :wolf:

I've just had my first chance in months to do a little diving into Polymer and wanted to get an app up and running with the Polymer Starter Kit and Firebase Hosting. Firebase Hosting is migrating to using [Superstatic](https://github.com/firebase/superstatic) as its serving engine (see [this PR for progress on that](https://github.com/firebase/superstatic/pull/187)) which means it'll be possible to run things identically locally.

I would **love** to make Superstatic work super-well with PSK so that PSK apps are dead simple to deploy to Firebase. There are a few catches though:

1. It gets really messy trying to make all of the relative URLs work with pushState.
2. Mounting `bower_components` as a route means Superstatic can't work its mojo.

What I had to do to make things work the way I needed:

1. Replace all relative links with absolute links in `app/index.html`
2. Move `bower_components` into `app`, `.gitignore` it there, and add a `.bowerrc` that points to `app/bower_components`
3. Update the `gulpfile.js` to install Superstatic as a middleware instead of the existing connect fallback middleware

None of this is terribly difficult, but it will be quite a few manual steps for Firebase users to go through to get PSK up and running. I'd very much prefer not to maintain a fork of PSK for Firebase's purposes, so I'm curious: are the above changes something that could potentially be put into the main repo? I'd be happy to pull it into a PR, but I'd want to know that such a PR has a chance of merge before I do the legwork here.

Goooooo Polymer!

MeTaNoV:
FYI, the is a Polymer recipe incoming in [PR 509](https://github.com/PolymerElements/polymer-starter-kit/pull/509) that you could follow and eventually complement/discuss.
I am also looking forward to this one!
Since you are also from Firebase, I would like to know if you have a pointer to show how to integrate a third party authentification? In my use-case, I would like to be able to sign-in use Steam authentification.

mbleigh:
@MeTaNoV it's definitely possible, but would require you to run your own server. Basically you'd need to redirect your users to a server you run that is capable of OAuth-ing to Steam, then you'd want to [mint a Custom Authentication Token](https://www.firebase.com/docs/web/guide/login/custom.html) for use with your Firebase app.

MeTaNoV:
@mbleigh , that's what I feared... I would have to do it GoogleAppEngine, but if I start there, I will tend to use the all its infrastructure...

chuckh:
[...]

PR #614, opened 2015-12-19
Use Superstatic + firebase.json for URL rewriting

mbleigh:
This PR makes some changes likely to have varying degrees of controversy:

1. Use Superstatic as the static web server for URL rewrites
2. Add a `firebase.json` (without app name) to the root (this is picked up by Superstatic)
3. Prefer absolute URLs in `index.html` over `<base>` tag
4. Default to pushState routing over hashbangs
5. Update Firebase deploy guide based on changes

Feedback? kthxbai :cat:

### DIFF
--- a/app/elements/my-greeting/my-greeting.html
+++ b/app/elements/my-greeting/my-greeting.html
@@ -26,7 +26,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   <script>
     (function() {
       'use strict';
-      
+
       Polymer({
         is: 'my-greeting',
 
--- a/app/elements/routing.html
+++ b/app/elements/routing.html
@@ -65,9 +65,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       page.redirect(app.baseUrl);
     });
 
-    // add #! before urls
     page({
-      hashbang: true
+      // add #! before urls for 
+      // hashbang: true
     });
 
   });
--- a/app/index.html
+++ b/app/index.html
@@ -12,6 +12,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <html lang="en">
 
 <head>
+  <base href="/">
   <meta charset="utf-8">
   <meta name="description" content="">
   <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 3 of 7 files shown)


robdodson:

robdodson:
This seems awesome to me. Local pretty URLs is a nice touch.

Some things to think about:

- If someone deploys to GitHub pages, or any other server that's not configured to always serve index.html they'll need to flip hashbangs back on and we'd need to document that.
- We could remove the firebase.json added to the `docs` directory by https://github.com/PolymerElements/polymer-starter-kit/pull/584
- The absolute URLs in `<head>` are tricky for folks serving from a subdirectory (like GH Pages). The docs for `<base>` say it only works for relative URLs. (I spoke with @mbleigh, we agreed to try to add a `<base>` element instead so it's easier to update our recipes)

mbleigh:
Switched to `<base>` but I'd argue in favor of keeping the `firebase.json` -- for one, it's used by Superstatic to pick up the configuration for rewrites. For another, it means that people just have to add one line and type `firebase deploy` to get it up on Firebase.

Switching to pushState by default will mean needing documentation for deploying to non-pushstate-able servers. I think pushState should be the default, though (esp. since hashbangs are considered deprecated by Google search crawlers).

robdodson:
> I'd argue in favor of keeping the firebase.json -- for one, it's used by Superstatic to pick up the configuration for rewrites. For another, it means that people just have to add one line and type firebase deploy to get it up on Firebase.

Oh yeah we'd keep the `firebase.json` you added. I meant we wouldn't include a 2nd one in the docs directory. Does that make sense?

mbleigh:
Yes, totally!

robdodson:
> Switching to pushState by default will mean needing documentation for deploying to non-pushstate-able servers. I think pushState should be the default, though (esp. since hashbangs are considered deprecated by Google search crawlers).

Yeah agreed. I'm all for getting folks on the right path

samccone:
@robdodson do you think this PR increasing the barrier to entry or decreasing it?

robdodson:
I don't see it as increasing the barrier. A significant portion of devs I talk to are already using PSK with Firebase so that part gets easier. For folks who aren't, it demonstrates how to do a SPA that doesn't rely on hashbangs which are a deprecated pattern at this point.

I do think it probably warrants a major version number rev though. But my hope is that for most folks this is a transparent change.

samccone:
Word, the push state routing default  was the big concern for me just because it does add a certain level of magic to things :)

chuckh:
I agree with moving to pushState by default. We will also need to update:

1. travis-runner.sh to handle this for the auto deploy to GitHub pages
2. Deploy to Github Pages https://github.com/PolymerElements/polymer-starter-kit/blob/master/docs/deploy-to-github-pages.md

robdodson:
@addyosmani are you around this week to chat about this?

addyosmani:
@robdodson I am! 

> Switching to pushState by default will mean needing documentation for deploying to non-pushstate-able servers. I think pushState should be the default, though (esp. since hashbangs are considered deprecated by Google search crawlers).

:+1:

samccone:
@addyosmani let me know if there is anything I can do here to help :wave:

mbleigh:
Likewise me :wave: :wolf:

addyosmani:
1. Will pushState by default (with it's magic) likely confuse users who don't have push-stateable servers? (Seems yes)
2. Are we content with moving ahead with it as a default assuming we can get a PR for docs for the non-pushState use case? - anyone want to write those docs? :)

robdodson:
> Will pushState by default (with it's magic) likely confuse users who don't have push-stateable servers? (Seems yes)

In particular this is the GitHub pages crowd. But page.js hash routing is super buggy so we're not in an awesome place there either. @mbleigh does super static support hash urls as an option?

> Are we content with moving ahead with it as a default assuming we can get a PR for docs for the non-pushState use case? - anyone want to write those docs? :)

Yeah if there's an easy enough fallback (maybe just swapping back to the old connect server and adding `hashbang: true` to the `page()` call in `routing.html`

mbleigh:
Since hash routes are all client side no reason superstatic can't support.
You'd just want to remove the rewrite from the config.

chuckh:
As a reminder the follow will also need to be updated for auto deploys to work after this pr is merged.

1. travis-runner.sh to handle this for the auto deploy to GitHub pages at https://github.com/PolymerElements/polymer-starter-kit/blob/master/travis-runner.sh#L13
 1. will need to set `hashbang: true`
 2. then test and make other possible fixes
3.  travis-runner.sh to handle this for the auto deploy to Firebase at https://github.com/PolymerElements/polymer-starter-kit/blob/master/travis-runner.sh#L26
 1. remove lines https://github.com/PolymerElements/polymer-starter-kit/blob/master/travis-runner.sh#L30-L31
 2. @mbleigh will this line still be needed? https://github.com/PolymerElements/polymer-starter-kit/blob/master/travis-runner.sh#L32

Also the Deploy to Github Pages recipe:  https://github.com/PolymerElements/polymer-starter-kit/blob/master/docs/deploy-to-github-pages.md will need to be updated to reflect these changes.

mbleigh:
@chuckh quite possibly. There is also official support for Firebase deploy baked into Travis now. See [this .travis.yml](https://github.com/staticapps/staticapps.org/blob/master/.travis.yml) for an example of the format needed.

abdonrd:
Any news?

robdodson:
I think the Polymer CLI is going to use polyserve as its local server. I'm wondering, with that in mind, if it makes sense to include a server at all with PSK2?

I would like to be able to support pretty urls instead of having to use hash routes. @justinfagnani is something like that in the cards for polyserver/Polymer CLI?

justinfagnani:
@robdodson definitely thinking about how to support pretty URLs in polyserve/polytool. No definite plans yet though.

robdodson:
Now that we've switched over to PSK2 and relying on Polymer CLI as our server, I think I'm going to close this and recommend that we build a separate generator, specifically targeted at Firebase + Polymer integration. This would also give us an opportunity to really show off Polymerfire and a bunch of the other Firebase features. Explained here as well https://github.com/PolymerElements/polymer-starter-kit/issues/512#issuecomment-242927182

mbleigh:
SGTM

Issue #279, opened 2015-07-26
Design issue


DmitryAuine:
For example if we use dom=shadow, styles of left-bar are diffrent

addyosmani:
To help us understand what's breaking here, can you provide some more details? For example, screenshots of what styles you're referring to in the left hand bar are different?

arthurvr:
Friendly ping @DmitryAuine.

DmitryAuine:
i am to lazy. U can just enable shadow and you will see that left menu has different style from shady

robdodson:
The issue I saw was that the color and background for selected items was not correct.

**With shady dom:**

<img width="517" alt="screen shot 2015-12-19 at 7 50 35 am" src="https://cloud.githubusercontent.com/assets/1066253/11913891/e16e4902-a627-11e5-9497-6065cf00e6b4.png">

**With shadow dom:**

<img width="492" alt="screen shot 2015-12-19 at 7 50 47 am" src="https://cloud.githubusercontent.com/assets/1066253/11913894/e56103b0-a627-11e5-98cf-b8589e21dd96.png">

Should be fixed by #617

PR #617, opened 2015-12-19
Use paper-menu mixin for selected styles. Fixes Gh-279

robdodson:
Addresses a small quirk referenced in #279.

Although the `::after` element uses `background: currentcolor` it produces a weird flicker (I guess something to do with the style shimmer) so I'm setting its value as well.


chuckh:
:+1:

PR #637, opened 2016-01-05
Update dependencies on README.me

### LINKED ISSUE #633: Python Dependency Issue for NPM install to complete.
The `node-gyp` dependency causes failure when using `npm install` if Python is not installed.

## Expected outcome

After running `npm install` and `bower update`, `gulp` + `gulp serve` should work.

## Actual outcome

An error is thrown in the middle of running `npm install` because of `node-gyp`. Specifically, because of python. 

## Steps to reproduce

1. Not having Python installed
2. Not having the correct version of python for `node-gyp`: [2.5, 3.0) 3.+ not supported.
3. Not having Python linked in environment variables, windows specific, or usable from cmd 
4. running npm install and normal procedures
5. Run gulp commands and errors will prevent it from performing.

## Note
1. Having Python 3 and above will cause problems with gulp.
2. Even after having python 2.5+ installed, the install gave warnings about path names being too long - but was able to complete installation.

## Fix
For windows
1. Install latest 2.X version of Python from https://www.python.org/downloads/
2. Place python in path http://stackoverflow.com/questions/6318156/adding-python-path-on-windows-7
3. Run npm install and bower update
4. Gulp should work if no other problems existed prior-hand.

Lemony-Andrew:
(For a single issue)
Fixes #633 

This fixes #633 by inserting necessary chronological dependency instructions.

- Added a definition for GYP
- Added a dependency step for gyp.

Further details: 
Important dependency instructions have been left out from the README.me. The node module `node-gyp` is included in this project; trying to run npm-install in the project will crash some way through unless its dependencies are handled prior-hand. This will result in other modules the project is dependent on not to be installed and users will not be able to use `gulp` in their project, which is an important feature to the intermediate to advanced package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ The full starter kit requires the following major dependencies:
 - npm, the node package manager, installed with Node.js and used to install Node.js packages.
 - gulp, a Node.js-based build tool.
 - bower, a Node.js-based package manager used to install front-end packages (like Polymer).
+- node-gyp, a cross-platform command-line tool written in Node.js for compiling native addon modules for Node.js
 
 **To install dependencies:**
 
@@ -74,7 +75,14 @@ npm install -g gulp bower
 
 This lets you run `gulp` and `bower` from the command line.
 
-4)  Install the starter kit's local `npm` and `bower` dependencies.
+4) Install node-gyp and its dependencies
+```sh
+npm install -g node-gyp
+```
+You may need to further install depencies for gyp varying by OS. Detailed instructions can be found on the gyp project: https://github.com/nodejs/node-gyp#installation
+
+
+5)  Install the starter kit's local `npm` and `bower` dependencies.
 
 ```sh
 cd polymer-starter-kit && npm install && bower install
--- a/README.md
+++ b/README.md
@@ -45,14 +45,16 @@ With Node.js installed, run the following one liner from the root of your Polyme
 npm install -g gulp bower && npm install && bower install
 ```
 
-#### Prerequisites (for everyone)
-
+#### Prerequisites 
+##### (for everyone)
 The full starter kit requires the following major dependencies:
-
 - Node.js, used to run JavaScript tools from the command line.
 - npm, the node package manager, installed with Node.js and used to install Node.js packages.
 - gulp, a Node.js-based build tool.
 - bower, a Node.js-based package manager used to install front-end packages (like Polymer).
+
+##### ( for Intermediate-Advanced )
+The Intermediate-Advanced starter kit uses node-gyp and requires the additional depency
 - node-gyp, a cross-platform command-line tool written in Node.js for compiling native addon modules for Node.js
 
 **To install dependencies:**
@@ -75,7 +77,7 @@ npm install -g gulp bower
 
 This lets you run `gulp` and `bower` from the command line.
 
-4) Install node-gyp and its dependencies
+4) (*Intermediate-advanced only*) Install node-gyp and its dependencies
 ```sh
 npm install -g node-gyp
 ```


Lemony-Andrew:
I Signed it!

README.md
```diff
@@ -45,15 +45,18 @@ With Node.js installed, run the following one liner from the root of your Polyme
...
+##### (for everyone)
 The full starter kit requires the following major dependencies:
-
 - Node.js, used to run JavaScript tools from the command line.
 - npm, the node package manager, installed with Node.js and used to install Node.js packages.
 - gulp, a Node.js-based build tool.
 - bower, a Node.js-based package manager used to install front-end packages (like Polymer).
 
+##### ( for Intermediate-Advanced )
+The Intermediate-Advanced starter kit uses node-gyp and requires the additional depency
```
samccone: this should be automatic... I am wondering how you got into this rabbit hole where you needed to manually install it?

Lemony-Andrew: It requires users to have Python. Windows 7 users like myself don't have Python pre-installed unless previously needed too. I have never worked with Python so I never installed it until after gyp. I can verify that it's not an automatic installation by my own experience, and the gyp-project requires you to install it by hand so it is not automated. 

I've read that people who work with Macs have Python basically pre-built in.

samccone:
Windows will always be a bit tricky to setup for node.. and if polymer starter kit is your first time using node it is going to be a rough time :(

This comes down to a failure in the node && npm community, and a failure of assumptions around who the users of node and npm will be.

https://twitter.com/luisrudge/status/601570367048654848
https://github.com/nodejs/node-gyp/issues?utf8=%E2%9C%93&q=windows
http://www.triplet.fi/blog/set-up-working-node-addons-node-gyp-environment-on-windows/
https://github.com/nodejs/node-gyp/issues/629

Unfortunately, I think adding a limited amount of documentation around node-gyp in our docs will cause more harm than help, perhaps a link and subheader for windows pointing at the blog post above I linked.

Lemony-Andrew:
Hi Samccone, I've been using npm and node for about 7 months, I've never actually experienced this problem before. This project was the first time I have used gulp though; bower and yo I've used in the past. 

I agree Polymer starter kit is well above beginner level for anyone starting out with node-like development. 

So instead of the step-by step insertion I've added just have a sub-header somewhere with the dependencies for windows? That would be helpful for anyone who's reading the instructions like I was expecting it go more smoothly then it ended up going. 

I'd prefer to add it somewhere before the npm-install step because that particular step took me at least 10-15 minutes to install each time trying to see what my problem was.

README.md
```diff
@@ -45,15 +45,18 @@ With Node.js installed, run the following one liner from the root of your Polyme
 npm install -g gulp bower && npm install && bower install
 ```
 
-#### Prerequisites (for everyone)
-
+#### Prerequisites 
+##### (for everyone)
 The full starter kit requires the following major dependencies:
-
 - Node.js, used to run JavaScript tools from the command line.
```
robdodson: I'd suggest we modify this to:

> - Node.js, used to run JavaScript tools from the command line. (Windows users will also need [node-gyp](http://www.triplet.fi/blog/set-up-working-node-addons-node-gyp-environment-on-windows/))

@samccone do you happen to know which module uses node-gyp?

robdodson:
@samccone @Lemony-Andrew

What do you think of updating this with some wording suggested by @tseelbach in https://github.com/PolymerElements/polymer-starter-kit/issues/642#issuecomment-170381724

addyosmani:
Friendly ping on this PR wrt the questions from Rob 😗

samccone:
I have gone ahead and added the note from #642 which I think does a good job of deferring responsibility  of the docs to node-gyp. Thanks for kicking this discussion into overdrive @Lemony-Andrew :)

Issue #583, opened 2015-12-11
Add auto deploy and test for Firebase


chuckh:
Currently I manually deploy to Firebase using below script. Lets add this process and possibly a test to travis_runner.sh.  I am working a pr now for this.

## Background:
@robdodson 
> @samccone yeah, right now it sounds like chuck is manually running these scripts. Even for the convenience of being able to always hit the URL and know it's running the latest stable PSK on Firebase would be useful.

@samccone
>sure we can add a CI deploy "test" that would then hit the server after the deploy and run a quick smoke test of the page via selenium....

## Script I created to deploy PSK to Firebase

```
clear
echo Build and Deploy PSK to Firebaseapp.com Server
echo  Comment: $1
echo
echo Making Changes to PSK for Firebase
cp firebase.json polymer-starter-kit/firebase.json
cd polymer-starter-kit
sed -i.tmp 's/<!-- Chrome for Android theme color -->/<base href="\/">\'$'\n<!-- Chrome for Android theme color -->/g' app/index.html
sed -i.tmp "s/hashbang: true/hashbang: false/" app/elements/routing.html

echo Starting Build Process
gulp
echo Finished Build Process, will start Firebase Deploy to PRODUCTION Server next
echo
echo Press any key to continue,  or control-c to abort
read -n 1 -s
echo
echo Starting Deploy Process to Firebaseapp.com Server -- polymer-starter-kit.firebaseapp.com
firebase deploy -m "$1"

echo Undoing Changes to PSK for Firebase
cp app/index.html.tmp app/index.html
cp app/elements/routing.html.tmp app/elements/routing.html
rm app/elements/routing.html.tmp
rm app/index.html.tmp
rm firebase.json

echo Done
firebase open

cd ../
```

robdodson:
Going to close this in favor of the PR if that's ok?

chuckh:
:+1:

PR #641, opened 2016-01-07
Add Auto Deploy to Firebase

chuckh:
Replaces pr #584 and addresses auto deploy part of issue #583.  PTAL @robdodson.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_script:
   - 'export DISPLAY=:99.0'
   - sh -e /etc/init.d/xvfb start
   - sleep 3
-  - npm i -g bower gulp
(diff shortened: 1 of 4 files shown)


robdodson:
annoying question... does the json file need to live in /docs or could it live in the directory with the firebase recipe?

chuckh:
Well... the recipe `deploy-to-firebase-pretty-urls.md` does live in the /docs directory.  We moved it there so it would not interfere with a new user wanting to use Firebase and running `firebase init`.

robdodson:
ohhh! my bad. I thought each recipe had its own directory within /docs (now i kind of wish they did so if there are extra files they could live with the recipe).

robdodson:
Not a blocker for this though. It LGTM, I'll let @samccone make the final call as he's more familiar with the travis stuff

addyosmani:
LGTM

Issue #596, opened 2015-12-15
Setting a behavior


klarkc:
Hello, I am having a problem here, and I can't figure why, I have this behavior, in the path: app/elements/behaviors/youtube-api-behavior.html

```
<script>
  FicBehaviors.YoutubeApiBehavior = {
    properties: {
      api: {
        type: Object
      },
      conf: {
        type: Object
      }
    },

    ready: function() {
      if (this.api) {
        // Execute callback
        this.readyApi(this.api, this.conf);
      }
    }
  };
</script>
```

This behavior is being loaded in the app/elements.html file and in the elements that use the behavior.
The FicBehaviors namespace is defined into the `<head>` tag, in app/index.html file, before the elements importation:

```
  <!-- build:js scripts/fic-behaviors.js -->
  <script src="scripts/fic-behaviors.js"></script>
  <!-- endbuild -->
```

fic-behaviors.js file:
```
// Define Behaviors Namespace
window.FicBehaviors = window.FicBehaviors || {};
```

According Polymer [docs](https://www.polymer-project.org/1.0/docs/devguide/behaviors.html), We can define a namespace in the window variable.

When I lint the project:
```
$ gulp lint
[12:30:32] Using gulpfile ~/psk/gulpfile.js
[12:30:32] Starting 'lint'...

app/elements/behaviors/youtube-api-behavior.html-script-0
  line 2  col 3  'FicBehaviors' is not defined.

  ⚠  1 warning


app/elements/fic-featured/fic-featured.html-script-0
  line 7  col 21  'FicBehaviors' is not defined.

  ⚠  1 warning

[12:30:34] 'lint' errored after 1.52 s
[12:30:34] Error in plugin 'gulp-jshint'
Message:
    JSHint failed for: app/elements/behaviors/youtube-api-behavior.html-script-0, app/elements/fic-featured/fic-featured.html-script-0
```

If I define a local FicBehaviors variable `var FicBehaviors = window.FicBehaviors`I have this error in lint:
```
app/elements/behaviors/youtube-api-behavior.html-script-0
  line 3  col 0  Expected indentation of 0 characters

  ⚠  1 warning
```

robdodson:
You need to add it to your `.jshintrc` file in the globals section: https://github.com/PolymerElements/polymer-starter-kit/blob/master/.jshintrc#L16-L22

Also you may want to have your default properties return new, empty objects:

```
properties: {
  api: {
    type: Object,
    value: function() {
[...]

Issue #531, opened 2015-11-23
jshint refers to *.html-script-0


zboralski:
jshint output from gulp serve refers to *.html-script-0.

zboralski:
The line numbers are useless. jshint should be ran on the original files instead.

samccone:
:+1: I agree this does seem bad, can you post the full error and how to reproduce?

jparish3:
Not sure if related, but now getting the following error on element used in prior starter kit versions (pre-1.2)  without error:
[15:19:25] 'lint' errored after 2.03 s
[15:19:25] Error in plugin 'gulp-jshint'
Message:
    JSHint failed for: app/elements/test-search/test-search.html-script-0

robdodson:
> The line numbers are useless. jshint should be ran on the original files instead.

hmm, I thought it was running only against [the original files](https://github.com/PolymerElements/polymer-starter-kit/blob/master/gulpfile.js#L110-L115)

quick question: are either of you using external javascript files for your elements?

robdodson:
I just ran into this while using the ES2015 recipe. Are either of you trying to do that or are you hitting this error on just a vanilla version of the app?

cnbuff410:
I tried ES2015 recipe and ran into this issue as well.

jparish3:
I am adding a slightly modified version of this element: https://github.com/kunalkundaje/address-autocomplete from the app/elements folder. it does not appear to be a bower package, so bringing it in manually

jesusabarca:
I just got that same error after updating to 1.2.1

```
[12:35:51] Error in plugin 'gulp-jshint'
Message:
    JSHint failed for: app/scripts/app.js, app/elements/house-data/house-data.html-script-0, app/elements/house-markers/house-markers.html-script-0, app/elements/log-in/log-in.html-script-0
[12:35:51] gulp-imagemin: Minified 6 images (saved 494 B - 1.4%)
```

robdodson:
I'm guessing one possible solution would be to tell jshint to ignore *.html-script-* files but I'd like to know what's creating them :D

jesusabarca:
@robdodson I'm new on this, so I don't really know how to troubleshoot it. Any suggestions would be much appreciated ;)

robdodson:
@jesusabarca in the short term just disable the lint task (remove it from being called in default or serve or anywhere else). I'll work on a fix tomorrow

samccone:
[...]

Issue #506, opened 2015-11-12
gulpfile gone nuts


gedw99:
cant work out whats wrong with gulp.

Env:
osx latest
node 5.0.0 (npm 3.3.6)

gulp is installed global.

```

gedw99-MacBook-Pro:static apple$ gulp
[16:19:22] Using gulpfile ~/workspace/go/src/bitbucket.org/gedw99/berlininfusions/server/static/gulpfile.js
[16:19:22] Starting 'clean'...
[16:19:23] Finished 'clean' after 311 ms
[16:19:23] Starting 'default'...
[16:19:23] Starting 'copy'...
[16:19:23] Starting 'styles'...
[16:19:24] Finished 'styles' after 970 ms
[16:19:25] copy all files 13.07 MB
[16:19:25] Finished 'copy' after 2.16 s
[16:19:25] Starting 'elements'...
[16:19:25] Finished 'elements' after 6.69 ms
[16:19:25] Starting 'jshint'...
[16:19:25] Starting 'images'...
[16:19:26] Starting 'fonts'...
[16:19:26] Starting 'html'...
[16:19:27] Finished 'fonts' after 692 ms

gulpfile.js
  line 1    col 1   Use the function form of "use strict".
  line 4    col 12  'require' is not defined.
  line 5    col 9   'require' is not defined.
  line 6    col 11  'require' is not defined.
  line 7    col 19  'require' is not defined.
  line 8    col 19  'require' is not defined.
  line 10   col 13  'require' is not defined.
  line 11   col 12  'require' is not defined.
  line 12   col 10  'require' is not defined.
  line 13   col 12  'require' is not defined.
  line 14   col 26  'require' is not defined.
  line 15   col 19  'require' is not defined.
  line 16   col 14  'require' is not defined.
  line 17   col 17  'require' is not defined.
  line 213  col 52  '__dirname' is not defined.
  line 310  col 1   'require' is not defined.
  line 314  col 5   'require' is not defined.

  ⚠  17 warnings


events.js:141
      throw er; // Unhandled 'error' event
      ^
Error: JSHint failed for: gulpfile.js

```

gedw99:
with node 5.0.0 the standard layout of the packages saved in the "node_modules" folder is different it seems. It has a flattened layout, rather than a tree.

gedw99:
SO, to test everything i pulled PSK 1.1.1

npm install, and 15 minutes later ( i am not joking). The mysterious node tells me:

````

[...]

PR #648, opened 2016-01-11
Remove linting

samccone:
People are way too confused by this. Going to back this out until we figure out a better strategy

https://github.com/PolymerElements/polymer-starter-kit/issues/506
https://github.com/PolymerElements/polymer-starter-kit/issues/531
https://github.com/PolymerElements/polymer-starter-kit/issues/596

(and many more)


robdodson:
SGTM. @addyosmani wdyt?

abdonrd:
A shame, I liked linting. :sob:

JosefJezek:
Unified linting is must have feature for developers on PSK ;-)

klarkc:
I like the liting feature, it's essential for the good quality of the code. But while there is not a clear way to refer directly the issue file, maybe the removal is the right thing to be done.

addyosmani:
I'm sad to see this go for now, but agree it's the right call. Let's explore how we can better redo linter support for the upcoming refactor.

samccone:
I am in agreement with you @abdonrd @JosefJezek @klarkc , however it seems like a large number of the users of PSK have never used a tool like jshint or jscs and are very confused about what is wrong with their code.

Due to the above fact, this generates a lot of noise and maintenance work for us on this repo. This is not a failure of our users, but rather a poor assumption on my part around how users will interact with our tooling.

robdodson:
For what it's worth, we're planning on redoing the build process so it's much easier to add in optional modules. Linting is going away for now but hopefully this is only temporary

robdodson:
@addyosmani I'm a bit confused. This PR was closed by e937a3d but that didn't remove linting unless perhaps the history got rewritten somewhere..? That commit is monstrous

felixzapata:
The problem with the lint process using JSCS is that the original task does not extract the script code like jshint does (jshint is more clear to show the line number of the error). 

Due to this reason, it was necessary to pass to a previous task to extract the script code (this task creates some temp files for each piece of script code).

PR #661, opened 2016-01-18
Remove jshint and jscs

robdodson:
Looks like https://github.com/PolymerElements/polymer-starter-kit/pull/648 somehow got overwritten and i'm getting complaints from folks using the latest PSK that they're hitting linter errors again. Must. Kill. JSHINT. @samccone @addyosmani PTAL


albertolobrano:
sad to see this goign .. :(

robdodson:
@albertolobrano we plan to bring it back in a way that's optional and easier for folks to setup

PR #671, opened 2016-01-22
Improve firebase doc fixes #670

### LINKED ISSUE #670: Suggestion: improve doc about Firebase Pretty URLs
I had a lot of problems when using Pretty URLs in Firebase, even following the guide: https://github.com/PolymerElements/polymer-starter-kit/blob/master/docs/deploy-to-firebase-pretty-urls.md

I discovered that the tag `<base href="/">` should come first in the `index.html`. When this tag was the last line of the `<head>` tag the browser had a lot of problems when finding the elements. But, if you put this tag just after the starting of the `<head>` and before any kind of import it works perfectly.

So, I would like to suggest changing the step text
`"Add <base href="/"> to head in index.html"`
to
`"Add <base href="/"> in the beggining of the head in index.html"`

chuckh:
With the travis-ci auto deploy to firebase we put `<base href="/">`
above `<!-- Chrome for Android theme color -->` It seems to work fine.

### DIFF
--- a/docs/deploy-to-firebase-pretty-urls.md
+++ b/docs/deploy-to-firebase-pretty-urls.md
@@ -48,7 +48,7 @@ guide](https://www.firebase.com/docs/hosting/quickstart.html).
           } ]
         }
 
-1.  Add `<base href="/">` to `head` in index.html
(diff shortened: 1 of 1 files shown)


samccone:
Thanks @chuckh :)

PR #675, opened 2016-01-23
Remove unneeded skip_files in the GAE deploy recipe

abdonrd:
These `skip_files` are unneeded because the deploy.sh uses the `dist` folder.


robdodson:
Thanks :)

PR #679, opened 2016-01-25
Trim down ensureFiles

robdodson:
@addyosmani @samccone PTAL

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -109,7 +109,7 @@ gulp.task('elements', function() {
 // "dot" files are specifically tricky due to them being hidden on
 // some systems.
 gulp.task('ensureFiles', function(cb) {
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -271,7 +271,7 @@ gulp.task('serve:dist', ['default'], function() {
 gulp.task('default', ['clean'], function(cb) {
   // Uncomment 'cache-config' if you are going to use service workers.
   runSequence(
-    ['copy', 'styles'],
+    ['ensureFiles', 'copy', 'styles'],
```
samccone: do we want this in any other step?

robdodson: can't think of any off the top of my head

samccone: kk

PR #684, opened 2016-01-28
Set Main Toolbar Color to White

chuckh:
The main toolbar was show black text and icons, instead of white.

See https://polymerelements.github.io/polymer-starter-kit/ and look at
title text.

### DIFF
--- a/app/styles/app-theme.html
+++ b/app/styles/app-theme.html
@@ -105,6 +105,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     padding: 0 16px;
   }
 
+  #mainToolbar {
+    color: #fff;
+  }
+
   #mainToolbar.tall .app-name {
     font-size: 40px;
     font-weight: 300;


samccone:
:+1:

abdonrd:
:+1:

robdodson:
Hm it didn't always do this. Did we regress somewhere?

samccone:
something must have

chuckh:
I think it has to do with this change on paper-toolbar: https://github.com/PolymerElements/paper-toolbar/commit/a702512966305e77e294c38769f58cb27483429a#diff-c039fb2eac608d42d39d9165a1afba13R114

You can see in paper-toolbar demo they are now setting paper-toolbar color to white at: 
https://github.com/PolymerElements/paper-toolbar/commit/58d1d0c0a0f5053c312bd2c2bfe2a13105a7cfbe#diff-3e22c50371c76d64ea9738dbc46a6e57R34

robdodson:
ah... https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/styles/app-theme.html#L28

robdodson:
cool thanks for tracking that down @chuckh !

samccone:
seems like a breaking change... perhaps we should file a bug upstream... @chuckh want to do that?

chuckh:
An issue for this problem has been open at: https://github.com/PolymerElements/paper-toolbar/issues/77.  See comment https://github.com/PolymerElements/paper-toolbar/issues/77#issuecomment-176499522.

PR #696, opened 2016-02-13
Update required node version in package.json

nelsonsilva:
According to the documentation the node version should be at or above 0.12.x.
This just updates the `package.json` accordingly since building with `0.10.x` does not work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,6 @@
     "start": "gulp serve"
   },
   "engines": {
-    "node": ">=0.10.0"
+    "node": ">=0.12.0"
   }
 }


robdodson:
Thanks!

samccone:
might be worth looking into what is required to get this working on 0.10

nelsonsilva:
@samccone you'd basically need to load a Promise polyfill for one of gulp's plugins but I'm guessing 0.10 will be impossible to support very soon with more and more plugins relying on ES6 support.
You can have a quick look at the errors by building it with nvm or Docker. I just used something like
`docker run -v "$PWD":/usr/src/app -w /usr/src/app node:0.10 /bin/bash -c "npm install -g gulp bower && npm install && bower install --allow-root  && gulp"`

Issue #340, opened 2015-08-28
Multi-page setup with neon-animation transitions


AaronLayton:
Provide documentation on how to setup a mutli-page app with `neon-animations`, possibly focusing on 

* Best practice way of splitting up your app to separate pages
* `hero-transition`
* `cascaded-animation`
* `slide-up-animation`

addyosmani:
I'll do my best to publish a PSK + neon-animation sample in the coming weeks that we can build on here. Alternatively might make for a good recipe.

stephenkeep:
@addyosmani have you had chance to work on this. Im looking to start a PSK project with animation this week.

samccone:
I know @ebidel may have been working on something related

CC @ebidel

robdodson:
FWIW, there are guides on [using neon-animations](https://elements.polymer-project.org/guides/using-neon-animations) in the Element's Catalog and I shot some polycasts covering using neon-animated-pages and the neon-animation behaviors.

- https://www.youtube.com/watch?v=wMhq1o0DULM&index=14&list=PLNYkxOF6rcIDdS7HWIC_BYRunV6MHs5xo
- https://www.youtube.com/watch?v=Lwvi1u4XXzc&list=PLNYkxOF6rcIDdS7HWIC_BYRunV6MHs5xo&index=16

MarcosGalaviz:
Rob very interesting, but how can we implement it with the sections of the Polymer Starter Kit? I am, a little confused :(

MarcosGalaviz:
I realy would like an example of PSK with neon-pages :(

JustinXinLiu:
I created a demo project ([code](https://github.com/JustinXinLiu/polymer-app-design) & [live demo](https://twitter.com/justinxinliu/status/685073908404371456)) here based on PSK with the use of `neon-animated-pages` instead of `iron-pages`.

Much credit should be given to @vguillou for his awesome [`neon-page-behavior`](https://github.com/vguillou/neon-page-behavior) add-on. It basically allows your page to use a different animation configurations when transitioning from/to other pages. I wish `neon-animated-pages` could have it built-in though!

P.S. Responsive design hasn't been done yet so please view it in Mobile device mode. :)

vguillou:
I had such an example ready so I published it : 
https://github.com/vguillou/polymer-starter-kit/tree/neon-animated-pages-basic

[...]

PR #700, opened 2016-02-17
New "Add page transitions with neon-animated-pages" recipe

vguillou:
#340


robdodson:
Looks awesome to me! Thanks for putting this together

PR #701, opened 2016-02-17
Change headers hierarchy to improve accessibility

felixzapata:
I've changed the levels of the headers in the examples due to accessibility. I think is better to use a level 1 header for the greetings component to not confuse the users in the examples.

I've used tools like [Claws](https://addons.mozilla.org/es/firefox/addon/claws/) or [Fangs](https://addons.mozilla.org/es/firefox/addon/fangs-screen-reader-emulator/) to review the headers levels.

### DIFF
--- a/app/elements/my-greeting/my-greeting.html
+++ b/app/elements/my-greeting/my-greeting.html
@@ -24,7 +24,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
     </style>
 
-    <h2 class="page-title">{{greeting}}</h2>
+    <h1 class="page-title">{{greeting}}</h1>
     <span>Update text to change the greeting.</span>
     <!-- Listens for "input" event and sets greeting to <input>.value -->
     <input value="{{greeting::input}}">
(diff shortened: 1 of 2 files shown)


robdodson:

robdodson:
Thanks!

I'm kind of wondering if the app name where it says 'Polymer Starter Kit' would be better served as an h1, while having each section heading be an h2 and then the legal heading be an h3 miiiiight be better but I'm not sure. I'll have to ask our local accessibility expert. Either way, having the legal heading be an h1 while all the section headings are h2s definitely doesn't make sense so thanks for this improvement!

felixzapata:
hi @robdodson, "my two cents" (like Kent Brockman):  set level 1 for "Polymer Starter Kit" I think is a valid option too, for this example of the starter. 

The tools I used are useful to decide the levels or to check if are correct or not.

The scenario can be a little different if you follow [HTML5 and the document outlining algorithm](https://www.smashingmagazine.com/2011/08/html5-and-the-document-outlining-algorithm/), but for the starter example it is not necessary (the levels are basic).

I hope my opinion won't be different from your local accessibility expert. 

By the way, and just for curiosity: [JAWS, IE and Headings in HTML5](http://accessibleculture.org/articles/2011/10/jaws-ie-and-headings-in-html5/)

Issue #699, opened 2016-02-16
node 0.10 support


samccone:
Let's investigate the work needed to get everything working on node 0.10

not supporting it is kinda rough for us atm

xref #696

addyosmani:
I know that we're going to need to support 0.12 for a while. Do we have any insight into how widespread 0.10 usage is atm? I really wish we could get more folks using the latest Node stable, but :crying_cat_face:

samccone:
Let me ask @seldo for the latest stats :)

PR #702, opened 2016-02-17
Fix node 0.10 support

nelsonsilva:
This should fix node 0.10 support (#699) since it includes a Promise polyfill required by at least PostCSS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ addons:
       - google-chrome-stable
       - g++-4.8
 node_js:
+  - '0.10'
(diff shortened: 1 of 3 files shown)


robdodson:
@samccone wdyt?

samccone:
lets add

0.12
0.10

here
https://github.com/PolymerElements/polymer-starter-kit/blob/master/.travis.yml#L12

and then :shipit:

samccone:
awesome, thanks @nelsonsilva :)

:clap:

Issue #704, opened 2016-02-18
Inform the user when navigate with the menu


felixzapata:
Hi, this issue is related with the accessibility. If a user clicks on the menu, the application should inform him that the page has changed maybe sending the focus, plus using the aria live regions.

I've made some tests but i am not sure which is the best approach to fix this. I mean, maybe this issue is more related with the iron-selectable behavior when it sets the "iron-selected" class. 

But maybe can this fix can be more related with the application itself, and when you uses the routing set the focus and the regions.  Or maybe it is necessary to add the iron-a11y-announcer in the starter.

Regards.

felixzapata:
I think that the solution can be like this way:

```javascript
    page('/users', function() {
      app.route = 'users';
      app.fire('iron-announce', { text: 'This is an announcement!' }, { bubbles: true });
    });
```

Maybe I will send a pull request.

robdodson:
I think the usual approach for managing focus in a situation like this is to set tabindex=-1 on one of the heading or section elements, then call focus on it after the page has transitioned

felixzapata:
yes, in the pull request I use the iron-a11y-announcer component to listen the event and to set the focus and the tabindex attribute.

PR #705, opened 2016-02-18
Use iron-a11y-announcer to inform the user when the menu has changed

felixzapata:
Related with the [issue 704](https://github.com/PolymerElements/polymer-starter-kit/issues/704).

With these changes, every time the user selects a menu option, the selected page will receive the focus and a message will be fired with the iron -a11y-announcer component.

What do you think?

### DIFF
--- a/app/elements/routing.html
+++ b/app/elements/routing.html
@@ -11,6 +11,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <script src="../bower_components/page/page.js"></script>
 <script>
   window.addEventListener('WebComponentsReady', function() {
+    
+    
+    window.Polymer.IronA11yAnnouncer.requestAvailability();
 
     // We use Page.js for routing. This is a Micro
     // client-side router inspired by the Express router
@@ -39,23 +42,28 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     page('/', function() {
       app.route = 'home';
+      app.fire('iron-announce', { text: 'Home section has been loaded', route:'home' }, { bubbles: true });
     });
 
     page(app.baseUrl, function() {
       app.route = 'home';
+      app.fire('iron-announce', { text: 'Home section has been loaded', route:'home' }, { bubbles: true });
     });
 
     page('/users', function() {
       app.route = 'users';
+      app.fire('iron-announce', { text: 'Users section has been loaded', route:'users' }, { bubbles: true });
     });
 
     page('/users/:name', function(data) {
       app.route = 'user-info';
       app.params = data.params;
+      app.fire('iron-announce', { text: 'User info section has been loaded', route:'user-info' }, { bubbles: true });
     });
 
     page('/contact', function() {
       app.route = 'contact';
+      app.fire('iron-announce', { text: 'Contact section has been loaded', route:'contact' }, { bubbles: true });
     });
 
     // 404
--- a/app/elements/routing.html
+++ b/app/elements/routing.html
@@ -12,9 +12,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <script>
   window.addEventListener('WebComponentsReady', function() {
     
-    
-    window.Polymer.IronA11yAnnouncer.requestAvailability();
-
     // We use Page.js for routing. This is a Micro
     // client-side router inspired by the Express router
     // More info: https://visionmedia.github.io/page.js/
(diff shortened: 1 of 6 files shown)


robdodson:
Did you have a chance to test with a screen reader? I'm curious if it is waiting to announce that the page has loaded before reading the focused content or if focusing the new section stops the current aria-live reading and starts reading the focused stuff immediately

felixzapata:
I've used ChromeVox extension because in the work I use Ubuntu and with Orca screen reader I can't check any webapge (I don't know why). 

It happens almost at the same time. With the ChromeVox extension you see that the section receives the focus and the voice anounce that the section has been loaded. The idea is to inform to the user that the page has changed its content.

robdodson:
yeah I understand the intent. I just want to make sure that it reads the content of the section and doesn't get stepped on by iron-a11-announce. You would much rather hear the heading for the section if you had to choose between the two. I can check in VoiceOver

robdodson:

felixzapata:
do you want that I change the focus to the header instead of the section?

robdodson:
I think focusing the section is fine. I think we need to add a style like section[tabindex=-1]:focus to `outline:none`. Otherwise there's a huge focus ring over the entire page.

Also I think we should remove the iron-a11y-announcer. In VoiceOver it is announcing _after_ the page content has been read, which is kind of weird.

felixzapata:
Well, due to accessibility is a good idea to add some kind of focus ... Maybe the default focus ring is too agressive but you can use for example a fade colour effect or similar.

And about the component, if you remove it, how do you know when to set the focus? I use this component for not to add the focus setting inside of the routing.html. I mean something like (it was my first attempt):

```javascript
    page('/users', function() {
      var target = document.querySelector('section[data-route="users"]');
      app.route = 'users';
      target.tabIndex = '-1';
      target.focus();
    });
```

robdodson:
> Well, due to accessibility is a good idea to add some kind of focus ... Maybe the default focus ring is too agressive but you can use for example a fade colour effect or similar.

I'm not sure if this is necessarily true in the case where you're managing focus. I spoke with a member of the Chrome accessibility team and this was their recommendation but I'll ask another friend in the accessibility community what the expectation would be.

> And about the component, if you remove it, how do you know when to set the focus? I use this component for not to add the focus setting inside of the routing.html. I mean something like (it was my first attempt):

Setting focus in the handler like that is fine. But you don't need to also announce a change as you're doing here (https://github.com/PolymerElements/polymer-starter-kit/pull/705/files#diff-6d2e463153e5eab168f79a9ac66557c6R45). Because that ends up getting read after the page has switched. Just setting focus when the route changes is enough because then the screen reader will begin announcing the content of that section.

felixzapata:
ok, I will change how to set the focus

felixzapata:
One example I forgot to mention before about the focus effect, you check [in this webpage](http://webaim.org/) how manage the effect when the elements receive the focus.

robdodson:
@felixzapata I'm not sure what you're asking?

felixzapata:
@robdodson mmm about the focus example? It is only an example that I not mentioned before where you can review one example of focus managing via some visual effect.

By the way, I´ve just update the pull request setting the focus in the routing.html file.

app/elements/routing.html
```diff
@@ -11,6 +11,9 @@
 <script src="../bower_components/page/page.js"></script>
 <script>
   window.addEventListener('WebComponentsReady', function() {
+    
+    
+    window.Polymer.IronA11yAnnouncer.requestAvailability();
```
robdodson: don't need this

app/scripts/app.js
```diff
@@ -14,7 +14,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   // and give it some initial binding values
   // Learn more about auto-binding templates at http://goo.gl/Dx1u2g
   var app = document.querySelector('#app');
-
+
```
robdodson: I would revert these white space changes

app/elements/routing.html
felixzapata: opps.. sorry

robdodson:
I think it would be good to give all of the section elements a `tabindex="-1"` then add a CSS style of `section[tabindex="-1"]:focus { outline: none; }`

app/scripts/app.js
felixzapata: do you mean the line blanks? I think that these lines are in the original app.js file in the master branch.. aren´t they?

felixzapata:
@robdodson I´ve updated the files but why the tests are failing?

robdodson:
Don't worry about the tests, it's probably just travis being weird. Thanks for the PR!

PR #707, opened 2016-02-19
Create recipe to add lint task to the starter

felixzapata:
In [current release the lint task was removed](https://github.com/PolymerElements/polymer-starter-kit/pull/661) but I propose this recipe in order to, meanwhile, give to the users an option to use _linters_ tools in the starter.

### DIFF
--- a/docs/add-lint-support.md
+++ b/docs/add-lint-support.md
@@ -1,0 +1,120 @@
+# Add lint support with JSHINT and JSCS
+
+This recipe helps you to create a task to use linters tools ([JSCS](http://jscs.info/) and [JSHINT](http://jshint.com/)) for your code. 
+
+
+## Create .jshintrc file in the root folder
+
+```
+{
+  "node": true,
+  "browser": true,
+  "bitwise": true,
+  "camelcase": true,
+  "curly": true,
+  "eqeqeq": true,
+  "immed": true,
+  "indent": 2,
+  "latedef": true,
+  "noarg": true,
+  "quotmark": "single",
+  "undef": true,
+  "unused": true,
+  "newcap": false,
+  "globals": {
+    "wrap": true,
+    "unwrap": true,
+    "Polymer": true,
+    "Platform": true,
+    "page": true,
+    "app": true
+  }
+}
+```
+
+## Create .jscsrc file in the root folder
+
+```
+{
+  "preset": "google",
+  "disallowSpacesInAnonymousFunctionExpression": null,
+  "disallowTrailingWhitespace": null,
+  "maximumLineLength": 100,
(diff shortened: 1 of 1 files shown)


chuckh:
We might want to make this recipe for eslint which Polymer is now using, see https://github.com/Polymer/polymer/pull/3382.

Also you will want to add a link to this recipe at the bottom of https://github.com/PolymerElements/polymer-starter-kit/edit/master/docs/README.md.

felixzapata:
I know that Polymer is now using eslint. My idea was that if this recipe was approved, then I would check the eslint too. I mean, recovering the original task as a recipe is quickly. This way, in the starter kit would exist an option to check the code.

docs/add-lint-support.md
```diff
@@ -0,0 +1,120 @@
+# Add lint support with JSHINT and JSCS
+
+This recipe helps you to create a task to use linters tools ([JSCS](http://jscs.info/) and [JSHINT](http://jshint.com/)) for your code.
```
robdodson: "linter tools"

docs/add-lint-support.md
```diff
@@ -0,0 +1,120 @@
...
+  "eqeqeq": true,
+  "immed": true,
+  "indent": 2,
+  "latedef": true,
+  "noarg": true,
+  "quotmark": "single",
+  "undef": true,
+  "unused": true,
+  "newcap": false,
+  "globals": {
```
robdodson: You only need `Polymer`, `page`, and `app`. The others are left over from Polymer <1.0

robdodson:
Won't this just get users back to where they were with the old linting solution? We received a lot of feedback that the reporter didn't do a good job of showing people where the errors were

docs/add-lint-support.md
felixzapata: done!

felixzapata:
yes, this task is a little confuse when throw the messages becuase of the auxiliar task to extract the script code inside the HTML. The original idea of this recipe was to recover the original task and add it as a recipe to those users who want use this kind of review.

But, after create the [other recipe with ESLint](https://github.com/PolymerElements/polymer-starter-kit/pull/708), maybe this one finally become obsolete and not neccessary.

robdodson:
Yeah the output from the ESLint task is actually really good! I would vote we discontinue work on JSHint and discourage its use in favor of ESLint because the reporter is so much better. Are you ok with me closing this PR?

felixzapata:
Yes, I agree. I like more the ESLint output because you avoid an extra package and there is no confuse with the errors lines.

abdonrd:
+1 to ESLint! :+1:

robdodson:
OK, closing this in favor of #708 then :)

PR #711, opened 2016-02-22
Remove unused task and watch

abdonrd:
From Polymer 1.1, the elements hasn't .css files.


robdodson:
yeah good call. I think we're recommending folks use "style modules" which come as HTML Imports in Polymer 1.0.

@samccone you ok with this?

abdonrd:
@robdodson exactly! Reference: https://blog.polymer-project.org/announcements/2015/08/13/1.1-release/

samccone:
yes :)

PR #712, opened 2016-02-22
Change the primary and accent colors from hex to vars

### LINKED ISSUE #709: app-theme colors with paper-styles/color
The primary-color and accent colors could be set by css variables from paper-styles/color.html

**Who will use the feature?**
Everyone that wants to change to a different, but material style conform color theme.

**What is the user’s goal?**
Change the theme quick, without needing to look at the Material Palette or the Material Design Color guidelines.

Given this style: 
```
:root {
    --primary-text-color: #212121;
    --primary-background-color: #ffffff;
    --secondary-text-color: #737373;
    --disabled-text-color: #9b9b9b;
    --divider-color: #b6b6b6;
    --primary-color: var(--paper-blue-500);
    --light-primary-color: var(--paper-blue-300);
    --dark-primary-color: var(--paper-blue-700);
    --accent-color: var(--paper-lime-a200);
    --light-accent-color: var(--paper-lime-a100);
    --dark-accent-color: var(--paper-lime-a400);
  }
```
A User can change to a red primary color and green accent color without much effort or any external sources.

**Downsides:**
Very new Users could be confused.

peter-mueller:
This fixes #709 by replacing the hex values with css variables to the colors from paper-styles/color.html.

closes #709

### DIFF
--- a/app/styles/app-theme.html
+++ b/app/styles/app-theme.html
@@ -10,6 +10,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 <link rel="import" href="../bower_components/polymer/polymer.html">
 <link rel="import" href="../bower_components/paper-styles/typography.html">
+<link rel="import" href="../bower_components/paper-styles/color.html">
 
 <style is="custom-style">
 
@@ -30,12 +31,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     --secondary-text-color: #737373;
     --disabled-text-color: #9b9b9b;
     --divider-color: #dbdbdb;
-    --primary-color: #3f51b5;
-    --light-primary-color: #c5cae9;
-    --dark-primary-color: #303f9f;
-    --accent-color: #ff4081;
-    --light-accent-color: #ff80ab;
-    --dark-accent-color: #f50057;
+    --primary-color: var(--paper-indigo-500);
+    --light-primary-color: var(--paper-indigo-100);
+    --dark-primary-color: var(--paper-indigo-700);
+    --accent-color: var(--paper-pink-a200);
+    --light-accent-color: var(--paper-pink-a100);
+    --dark-accent-color: var(--paper-pink-a400);
 
     /* Components */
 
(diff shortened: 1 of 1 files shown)


robdodson:
@notwaldorf if you have time would you mind reviewing this PR? I am still really iffy on my understanding of which elements implement paper-styles but am hoping that this will just work? :smiley:

samccone:
This will just work I am 99.9% sure @robdodson 

+1 thanks for the patch @peter-mueller :)

app/styles/app-theme.html
```diff
@@ -30,12 +31,12 @@
     --secondary-text-color: #737373;
     --disabled-text-color: #9b9b9b;
     --divider-color: #dbdbdb;
-    --primary-color: #3f51b5;
-    --light-primary-color: #c5cae9;
-    --dark-primary-color: #303f9f;
-    --accent-color: #ff4081;
-    --light-accent-color: #ff80ab;
-    --dark-accent-color: #f50057;
+    --primary-color: var(--paper-indigo-500);
```
abdonrd: --primary-text-color: var(--paper-grey-900);

abdonrd:
Yes, It works. LGTM except one comment. :)

peter-mueller:
I'm a bit unsure about the text and divider color, as [Material-Design Color Guide](https://www.google.com/design/spec/style/color.html#color-text-background-colors) states opacity should be used. (under *use opacity instead of grey*). I would like to change this part to using opacity, but if I understand correctly, almost every paper-element would have to be changed to support it.

notwaldorf:
I don't think we should be moving towards using opacity instead of grey. I am still trying to get someone from Blink to write down a definite answer (see https://github.com/PolymerElements/paper-styles/pull/98#issuecomment-185985819), but basically: painting transparent text is not optimized and will be more expensive, and we shouldn't do that just because the spec says so.

notwaldorf:
@robdodson Why don't you import `paper-styles/default-theme.html` and get a lot of those values for free? That's exactly what that file does...

samccone:
I think the motivation here was to explicitly define them at the top of the file so users can go in and quickly just change the values. :)

Protoss78:
Great PR. I think it will make it much easier to customize the application theme. Plus it will increase readability.

PR #715, opened 2016-02-22
rm jshint since we removed it in 1.2.3

dandv:


robdodson:
oh shoot I didn't realize that was still in there. Thanks!

PR #721, opened 2016-02-26
ignore dist folder

marvinrichter:
`node_modules/ ` and `bower_components` will be ignored by default like described in their [documentation](http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories).

Instead, ignoring the `dist/` makes sense because other than in the gulp task where we explictly define `app/*` a plugin in an editor has it's entry point in the root folder.

### DIFF
--- a/docs/add-eslint-support.md
+++ b/docs/add-eslint-support.md
@@ -31,8 +31,10 @@ This recipe helps you to create a task to use [ESLint](http://eslint.org/) tool.
 ## Create .eslintignore file in the root folder
 
 ```
-node_modules/*
-bower_components/*
+# /node_modules and /bower_components ignored by default
+
+# Ignore built files
+dist/
 ```
 
 


robdodson:
LGTM. @felixzapata does this sound OK to you?

abdonrd:
I prefer this:
```
dist/
```

Instead of this:
```
# /node_modules and /bower_components ignored by default

# Ignore built files
dist/
```

But LGTM!

samccone:
landing we can iterate more :)

awesome work @marvinrichter thanks for taking the time to improve this repo, your time and work is really appreciated!

marvinrichter:
You're welcome :)

felixzapata:
hi @robdodson, sorry for the delay. It is ok for me. I took the original file content of the [Polymer project repository](https://github.com/Polymer/polymer/blob/master/.eslintignore) and I did not know about the default configuration :)

Issue #724, opened 2016-02-26
Support for Google App Engine using custom domains and http only


johnwhitton:
Hi,
When deploying using the [app engine recipe](https://github.com/PolymerElements/polymer-starter-kit/issues/723) my custom domain stopped working because I had not set up SSL certificates for it on google app engine. To get around this I commented out  the `secure: always` lines from app.yaml you can see the change [here](https://github.com/johnwhitton/jincubator/commit/4289b2db59cd51964f51af15b5856ee5942b390c) 
I'm wondering if there is valued adding a comment to the recipe to mention this. Say something like
This is the configuration file for the GCP project. It sets a python runtime environment and static file handlers. The configuration utilizes GAE HTTP/2 capabilities in order to minimize load time for HTTP/2 compatible browsers. **​_Note this also ensures https, if you wish to use custom domains supporting http only you will need to remove all the 'secure: always’ entries._​**

Regards,

John

samccone:
hey @johnwhitton thanks for the comment, I think that making a note of this in the deploy process doc is a really good idea!

Any interest in opening a PR with this change? I will gladly merge :)

Thanks for taking the time to do research and write up your findings, it is super appreciated :sparkles: !

johnwhitton:
Hey @samccone sure

robdodson:
@johnwhitton btw if you server over http it means you can no longer use things like service worker or web app manifests. I haven't tried using it with app engine yet, but a service like [Let's Encrypt](https://letsencrypt.org/) can make it easier to get a cert.

johnwhitton:
@robdodson Yes HTTPS is something I should set up, there are a couple of blogs about using Let's Encrypt [here](http://blog.seafuj.com/lets-encrypt-on-google-app-engine) and [here](https://medium.com/@samthor/let-s-encrypt-with-app-engine-8047b0642895#.d1wkbpgr6). Also note that regardless of your custom domain in my case www.jincubator.com appspot supports HTTPS e.g. https://jincubator-ae.appspot.com/
It was just confusing for me when I first deployed and my I could not access www.jincubator.com, so I thought I'd mention it. If the team prefers to recommend against using HTTP or put additional warnings such as @lozandier mentions in [725](https://github.com/PolymerElements/polymer-starter-kit/pull/725) and [726](https://github.com/PolymerElements/polymer-starter-kit/pull/726) then that makes sense to me.

Regards,

John

PR #725, opened 2016-02-26
Update Google App Engine recipe to include http support

johnwhitton:
This pull request is to add additional documentation to the Google App Engine recipe and fixes https://github.com/PolymerElements/polymer-starter-kit/issues/724


samccone:
LGTM thanks again @johnwhitton

lozandier:
@samccone @johnwhitton Shouldn't it note that they're aware if they decide to have custom domains that are `http` only that some elements/APIs such as the platinum element that use the ServiceWorker won't work  (requiring `https`)?

PR #727, opened 2016-02-27
Update .eslintignore

abdonrd:
We need to ignore the `bower_components` because we have it in the app folder, and not in the root folder.

### DIFF
--- a/docs/add-eslint-support.md
+++ b/docs/add-eslint-support.md
@@ -31,10 +31,8 @@ This recipe helps you to create a task to use [ESLint](http://eslint.org/) tool.
 ## Create .eslintignore file in the root folder
 
 ```
-# /node_modules and /bower_components ignored by default
-
-# Ignore built files
-dist/
+/app/bower_components/**
+/dist/**
 ```
 
 


marvinrichter:
You're totally right. Because the included default pattern is `/bower_components` it is relative to the root folder.

Anyway, I wouldn't make the pattern absolute but a more general matching pattern like `bower_components` -> match any file or folder named bower_components.

`/dist/**`is a performance killer if you have many folder/files in the dist folder because for each file/folder it would check if it matches the ignore pattern. Instead just ignore `/dist` (any file or folder absolute from root) or `/dist/` (any folder absolute from root).

abdonrd:
Thanks for the explications, @marvinrichter!

I have proposed these changes after reading the following in the [ESLint documentation](http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories):

> In addition to any patterns in a .eslintignore file, ESLint always ignores files in `/node_modules/**` and `/bower_components/**`.

Then... The best option is this?
(I prefer make absolute pattern because in PSK the `bower_components` is always inside `app`)

```
/app/bower_components
/dist
```

marvinrichter:
Actually I have to take that back with the performance. I just checked their sources and they index the cwd anyway. So if it is easier to understand for one, then just leave it 
```
/app/bower_components/**
/dist/**
```

PR #729, opened 2016-03-01
Fix Markdown

dandv:


samccone:
:+1: nice thanks a ton @dandv :clap:

PR #731, opened 2016-03-01
Update watch app in add-eslint-support.md

abdonrd:


samccone:
:+1: thx good spot @abdonrd

Issue #687, opened 2016-01-31
Browser sync binding and CSP


zboralski:
After watching @robdodson's last polycast, I decided to enable CSP and use crisper to separate html and javascript code and the last road block is the browser-sync-binding snippet : 

```html
  //<![CDATA[
  document.write(
    "<script async src='/browser-sync/browser-sync-client.2.11.0.js'><\/script>"
    .replace("HOST", location.hostname));
  //]]>
```

I am not sure how I should go around to fix it.

robdodson:
Is that line of code coming out of the browser sync task itself?

addyosmani:
We have static browser-sync markup in PSK's index over at https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/index.html#L63 and I believe that's used for injection over in places like https://github.com/PolymerElements/polymer-starter-kit/blob/5929c0b16c170f6d8d5ac115167458652ac24161/gulpfile.js#L218. 

@zboralski If you drop https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/index.html#L63 entirely from your setup, do you run into any issues with crisper at all?

zboralski:
@addyosmani I removed the span which gets replaced by some inline code. It would help if the injection is done before vulcanise. Although I think vulcanize is only ran on elements.html.

zboralski:
@addyosmani if I remove https://github.com/PolymerElements/polymer-starter-kit/blob/master/app/index.html#L63 I have no issues.

samccone:
:+1: @zboralski seems like the correct resolution. I am glad you got it worked out. Hopefully this issue will act as a good reference if someone finds themselves in a similar situation. Thanks for digging in on this one.

zboralski:
@samccone why should the snippet be injected for the dist version ?  I don't think this is resolved correctly. I'd like to be able to use browser sync during development but it shouldn't interfere with the dist version.

samccone:
> I'd like to be able to use browser sync during development but it shouldn't interfere with the dist version.

^ This is an issue that you should only hit when using CSP

+1 to this goal however, this would be ideal, we would need to inject the browser-sync code dynamically during dev or just dynamically remove the chunk when building the final dist.

---

The reason I closed this is because at a certain point of customization of the PSK you will hit these kind of issues, and making a change to the code is to be expected, since this is just a starter kit :)

[...]

PR #732, opened 2016-03-01
#687 Remove `browser-sync` span in dist

vguillou:


robdodson:
oh neat, does this just magically work with useref?

vguillou:
Yep

(Somehow forgot to reference the issue, so here #687)

samccone:
:dancer: :dancer: :dancer: :dancer: :dancer: :dancer: :dancer: 

:hand: :five: @vguillou

Issue #718, opened 2016-02-24
Fixup ES2015 recipe


robdodson:
Currently the ES2015 recipe doesn't work for `gulp serve:dist`. Also related https://github.com/PolymerElements/polymer-starter-kit/issues/625

robdodson:
cc @nevraeka

Nevraeka:
thanks @robdodson . I'm looking into this tonight

buele:
1. Following the recipe: [Add ES2015 support through Babel](https://github.com/PolymerElements/polymer-starter-kit/blob/master/docs/add-es2015-support-babel.md)
 2. Running ```gulp serve:dist``` this error appears: 

```
$ gulp serve:dist
[15:32:42] Using gulpfile /data/buele/git/polymer-starter-kit/gulpfile.js
[15:32:42] Starting 'clean'...
[15:32:42] Finished 'clean' after 69 ms
[15:32:42] Starting 'default'...
[15:32:42] Starting 'ensureFiles'...
[15:32:42] Finished 'ensureFiles' after 419 μs
[15:32:42] Starting 'copy'...
[15:32:42] Starting 'styles'...
[15:32:43] styles all files 98 B
[15:32:43] Finished 'styles' after 755 ms
[15:32:43] copy all files 1.05 MB
[15:32:43] Finished 'copy' after 925 ms
[15:32:43] Starting 'elements'...
[15:32:43] Starting 'js'...
[15:32:43] 'js' errored after 1.45 ms
[15:32:43] TypeError: Cannot read property 'init' of undefined
    at Gulp.<anonymous> (/data/buele/git/polymer-starter-kit/gulpfile.js:124:23)
    at module.exports (/data/buele/git/polymer-starter-kit/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:34:7)
    at Gulp.Orchestrator._runTask (/data/buele/git/polymer-starter-kit/node_modules/gulp/node_modules/orchestrator/index.js:273:3)
    at Gulp.Orchestrator._runStep (/data/buele/git/polymer-starter-kit/node_modules/gulp/node_modules/orchestrator/index.js:214:10)
    at Gulp.Orchestrator.start (/data/buele/git/polymer-starter-kit/node_modules/gulp/node_modules/orchestrator/index.js:134:8)
    at runNextSet (/data/buele/git/polymer-starter-kit/node_modules/run-sequence/index.js:88:16)
[...]

PR #734, opened 2016-03-02
[#718] Fixed serve errors caused by ES6 transpilation

Nevraeka:
Description:
closes issue #718

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -104,6 +104,10 @@ gulp.task('styles', function() {
   return styleTask('styles', ['**/*.css']);
 });
 
+gulp.task('elements', function() {
+  return styleTask('elements', ['**/*.css']);
+});
+
 // Ensure that we are not missing required files for the project
 // "dot" files are specifically tricky due to them being hidden on
 // some systems.
@@ -173,6 +177,20 @@ gulp.task('vulcanize', function() {
     .pipe($.size({title: 'vulcanize'}));
 });
 
+// Transpile all JS to ES5.
+gulp.task('js', function () {
+  return gulp.src(['app/**/*.{js,html}', '!app/bower_components/**/*'])
+    .pipe($.sourcemaps.init())
+    // Extract JS from .html files
+    .pipe($.if('*.html', $.crisper({scriptInHead:false})))
+    .pipe($.if('*.js', $.babel({
+      presets: ['es2015']
+    })))
+    .pipe($.sourcemaps.write('.'))
+    .pipe(gulp.dest('.tmp/'))
+    .pipe(gulp.dest(dist()));
+});
+
 // Generate config data for the <sw-precache-cache> element.
 // This include a list of files that should be precached, as well as a (hopefully unique) cache
 // id that ensure that multiple PSK projects don't share the same Cache Storage.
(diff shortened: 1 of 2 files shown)


Nevraeka:
cc @samccone

gulpfile.js
```diff
@@ -104,6 +104,10 @@ gulp.task('styles', function() {
   return styleTask('styles', ['**/*.css']);
 });
 
+gulp.task('elements', function() {
```
robdodson: I think this task was removed in https://github.com/PolymerElements/polymer-starter-kit/commit/abf3309d87116eee85b5a00d8163ea3bbdfe37cc because we never ship external CSS files with our elements

gulpfile.js
```diff
@@ -173,6 +177,20 @@ gulp.task('vulcanize', function() {
     .pipe($.size({title: 'vulcanize'}));
 });
 
+// Transpile all JS to ES5.
+gulp.task('js', function () {
```
robdodson: I think we only want to do this in the recipe, not the primary gulpfile. Right @samccone ?

gulpfile.js
```diff
@@ -271,11 +319,11 @@ gulp.task('default', ['clean'], function(cb) {
   // Uncomment 'cache-config' if you are going to use service workers.
   runSequence(
     ['ensureFiles', 'copy', 'styles'],
+    ['elements', 'js'],
```
robdodson: don't think we need elements here

gulpfile.js
```diff
@@ -214,6 +232,36 @@ gulp.task('clean', function() {
...
+    // https: true,
+    server: {
+      baseDir: ['.tmp', 'app'],
+      middleware: [historyApiFallback()]
+    }
+  });
+
+  gulp.watch(['app/**/*.html'], ['js', reload]);
+  gulp.watch(['app/styles/**/*.css'], ['styles', reload]);
+  gulp.watch(['app/elements/**/*.css'], ['elements', reload]);
```
robdodson: can remove this line

gulpfile.js
```diff
@@ -300,9 +348,21 @@ gulp.task('deploy-gh-pages', function() {
...
+var injectDependencies = function(task, dependencies) {
+
+  if (task !== null) {
+    dependencies.forEach(function(dependency) {
+      task.dep.push(dependency);
+    });
+  }
+};
+
+var wctTestDependencies = gulp.tasks.serve.dep;
```
robdodson: Hm I thought WCT started its own local server. Are you not able to run tests with that?

gulpfile.js
```diff
@@ -300,9 +348,21 @@ gulp.task('deploy-gh-pages', function() {
...
+
+injectDependencies(gulp.tasks[ 'wct:local' ], wctTestDependencies);
+injectDependencies(gulp.tasks[ 'wct:sauce' ], wctTestDependencies);
```
robdodson: Do you need to run serve task to test on saucelabs?

robdodson:
Does this continue to work when you run `gulp serve:dist`?

Nevraeka:
It works for me locally with `gulp serve: dist`

gulpfile.js
Nevraeka: I'm not sure so I can remove this if we need

robdodson: Yeah it doesn't seem related to the ES2015 stuff

gulpfile.js
Nevraeka: ok. For some reason I thought the elements stuff was necessary but that makes sense

Nevraeka:
travis is breaking on testing nodejs v0.10 which I think we agreed is no longer a nodejs target for this project . Please verify @samccone . I can change the travis config to remove [this line](https://github.com/PolymerElements/polymer-starter-kit/blob/master/.travis.yml#L13) if we need the build to pass. Thoughts?

samccone:
We do not want the changes here to be in the gulpfile since this is part of a recipe right?
The updates should be made to the es2015 recipe instead I think :)

abdonrd:
+1 to @samccone! :)

Nevraeka:
@samccone @abdonrd sounds good. I'll close the PR if I can and add the changes to the recipe

PR #733, opened 2016-03-02
Add "Add to home screen" section. Fix #95 for real.

dandv:

Issue #95, opened 2015-06-01
Where's "Web App Install Banner"?


shawnExplore:
This was presented as a "feature" of the starter kit, but no such install banner pops up on my android device, nor can I find any code for this banner while browsing through this repo.

Do I need some specific version of browser for this to work, or is it not actually implemented? Also, I notice the app he's actually showing is IOWA, which also doesn't pop up any such banner when I navigate to the url from my Android Lollipop phone..

[video showing "web app install banner"](https://youtu.be/fD2As5RmM8Q?t=2210)

I'd love to know how to display such a banner  - it is an install experience almost as easy (easier) as the native play store.

deearcodes:
@shawnExplore There's no "Install Banner". Using a supported browser (I'm using Chrome) go to browser menu while browsing the Polymer Starter Kit web-app. If your ´manifest.json´ and ´<meta>´ tags are fine tap "Request desktop version". Search for the Polymer Starter Kit logo in your home screen.

shawnExplore:
With further investigation it appears this feature has nothing to do with Polymer at all, and in fact isn't __directly__ controllable by the developer. It's something [introduced in Chrome 42](https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android?hl=en).  Please correct me if I'm wrong, but highlighting this "feature" (see video link) in a Polymer talk was a bit of false advertising :-1:  Feel free to reopen this ticket if it is supposed to be a feature of this repository.

addyosmani:
@shawnExplore It's absolutely not false advertising. The native UI for the feature is exposed in Chrome 42, but requires several prerequisites normally manually setup by a developer in order to be enabled. This includes the web app manifest being required (we provide this), Service Worker setup (non trivial, and we provide this) and sufficient favicons setup. We also, provide this. Any normal website or webapp without these setup is not going to get access to the functionality.

dandv:
The requirements for the app install banner to show have fluctuated over time, which might explain why the feature hasn't worked for @shawnExplore.

I've updated our [docs](https://github.com/google/WebFundamentals/pull/2617) to mention the requirement for defining a `start_url` in `manifest.json`, and PR #733 adds a section about adding the setting.

addyosmani:
Thanks so much for updating the docs and landing this change for PSK, @dandv! We really appreciate it :)

PR #735, opened 2016-03-02
Mention caching of start_url when opening from home screen

sorin-davidoi:
Expands on #733, related to issue #95.

Since we have added a section on triggering a web app install banner, I feel we should mention this common pitfall and point the users in the right directions.


sorin-davidoi:
I signed it!

samccone:
oh snap, good call @sorin-davidoi :) :clap: thanks for the PR very much appreciated.

dandv:
:+1:

PR #749, opened 2016-03-15
Fix the user-info page paper-material elevation

vguillou:
It was recently set to -1 by mistake.


robdodson:
LGTM thanks. Must have been an accident when we were updating the accessibility code

samccone:
berp mah bad!

PR #750, opened 2016-03-16
Improve focus managment

robdodson:
@samccone PTAL


samccone:
moonrise-kingdom-600 78


LGTM

PR #753, opened 2016-03-17
Responsively hide drawer panel @840px

joryphillips:
The default responsive width to hide the paper-drawer-panel is an uncomfortable 600px -- let's increase it to 840px, a value consistent with media queries found in the PSK.


samccone:
oh snap :clap: @joryphillips

chuckh:
LGTM2.

samccone:
:boom:

PR #758, opened 2016-03-17
Fix #757

### LINKED ISSUE #757: Remove search and reload icon from header on mobile
They do nothing so we should just drop them :)

abdonrd:


samccone:
:+1: 

just going to wait for @addyosmani or @robdodson or @chuckh to also +1 this

Issue #653, opened 2016-01-12
V(n) goal posts 


samccone:
#### Below is a brief list of high level goals that I would like to really nail with the v(n) release :sparkles: 

* [x] Well defined (file directory) structure for components and apps
* [ ] Generator to quickly add new components and pages
* [x] Rock solid baked in router solution
* [ ] A Simple Unit testing approach (mocha / ava)
* [x] **An** opinionated build tool*
* [ ] Import / require javascript everywhere / anywhere
* [ ] State management --- redux wrapper? immutable js? -- we need page level and view level state management that just works 
* [ ] Simple deployment -- https://github.com/carrot/ship <--- or something like it

robdodson:
cc @tjsavage

abdonrd:
Sounds great! :grin:

addyosmani:
Thanks for sharing this list, @samccone! We have a week of Polymer planning for this quarter lined up and it would be great to look at how we can tackle some of these items as part of our high-level goals. I'd love to see us rewriting PSK for more flexible layout (app-layout elements), the new router (carbon-router) and what we can do to both simplify our tooling (before we add in more) and lower friction for progressive web apps.

> Generator for quickly adding new components and pages

We've been kicking around an idea for maybe doing a polymer-cli. This could be something that abstracts away the Yeoman generator but adds in support for polylint, polyclean and so on without you having to mentally reason what to use. It'll be interesting to hear what you and the rest of the team think of going in that direction.

> An opinionated build tool

What does this mean? :) My experience with polybuild (which may be what you're alluding to) tells me that it can be hard for us to ship something that caters for the numerous configurations folks want us to support, but that doesn't mean we shouldn't do it better. Will wait for you to expand on this.

> State management

Imo, it would be good to bring this up with the team to see if there's a Polymer-world view on state management that could be built or adopted. Remember that we'll also be trying to reduce the complexity overhead of using PSK and opting for Redux or an immutability library may mean we expect folks to learn how those work before they can really take advantage. Not at all saying no, simply saying we should keep that in mind.


> Import/require JavaScript everywhere

It would be useful to look at the tooling pipeline for what folks would expect here. Babel support everywhere + ES2015 linting (?) + anything else? It would be good to learn from the issues we ran into with support for ES2015 in current PSK. Are you also asking for CJS require support?

> Deployment

I'm +1 for deployment.

chuckh:
:+1:  for @samccone  list.

[...]

PR #769, opened 2016-04-03
PSK2

robdodson:
A lot of folks have been asking for carbon-route and app-layout integration. I figure now would be a good time to kickoff work on PSK2.

High level goals are here: https://github.com/PolymerElements/polymer-starter-kit/issues/653

Starting a PR just so it's easier to comment on stuff. This first commit is kinda crazy because it rips out most of the old PSK and replaces it with a stripped down shell similar to [Zuperkulblog (as seen in the app-layout demos)](http://polymerlabs.github.io/app-layout/).

Lots more to do but I have to go to sleep now :P

@samccone @chuckh @addyosmani @abdonrd @TimvdLippe


robdodson:
Oh I forgot to mention. To test the branch you'll want to serve with `polyserve`

abdonrd:
@robdodson why move the app content to the root folder?

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
+<!--
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
```
abdonrd: 2016? And the same with the other licenses.

index.html
```diff
@@ -6,20 +6,28 @@
...
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
+  <meta name="theme-color" content="#fafafa">
+
+  <title>STARTER KIT</title>
+
+  <link rel="manifest" href="manifest.json">
+
+  <script src="../webcomponentsjs/webcomponents-lite.js"></script>
```
abdonrd: How to deploy later with the relative paths?

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
...
+
+      ready: function() {
+        this.async(function() {
+          if (!this.route.path) {
+            this.set('route.path', '#/art');
+          }
+        });
+      },
+
+      _stringify: function(obj) {
```
abdonrd: Why this?

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
...
+          </app-toolbar>
+
+          <!-- bottom toolbar -->
+          <app-toolbar class="title-toolbar main-title-toolbar">
+            <div class="title">STARTER KIT</div>
+          </app-toolbar>
+
+        </app-header>
+
+      </app-header-layout>
```
abdonrd: Here something like this?

```html
<iron-pages selected="[[params.page]]" attr-for-selected="data-page">
  <psk-page-art data-page="art"></psk-page-art>
  <psk-page-film data-page="film"></psk-page-film>
  <psk-page-design data-page="design"></psk-page-design>
</iron-pages>
```

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
...
+      is: 'psk-app',
+
+      properties: {
+
+      },
+
+      ready: function() {
+        this.async(function() {
+          if (!this.route.path) {
+            this.set('route.path', '#/art');
```
abdonrd: This is not working me. :confused:

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
...
+
+          <!-- nav menu -->
+          <iron-selector
+              class="nav-menu"
+              selected="[[params.page]]"
+              attr-for-selected="data-page"
+              on-iron-activate="_drawerSelected">
+            <a data-page="art" href="#/art">Art</a>
+            <a data-page="film" href="#/film">Film</a>
+            <a data-page="design" href="#/design">Design</a>
```
abdonrd: I'm not sure if it is the best way, but... WDYT about use binding for the routes? Like this:

```html
<a data-page="art" href$="{{routes.art}}">Art</a>
<a data-page="film" href$="{{routes.film}}">Film</a>
<a data-page="design" href$="{{routes.design}}">Design</a>
```

abdonrd:
Nice work @robdodson!

I think that would be awesome if we implement an [`app-shell`](https://developers.google.com/web/updates/2015/11/app-shell) solution too.

TimvdLippe:
Initial version looks fine. Once we have a more concrete structure I will take a look at the routing and structure of the elements. :+1:

src/psk-app/psk-app.html
robdodson: Left over from debugging the router. I'll remove it

robdodson:
@abdonrd  asked:

> why move the app content to the root folder?

I think you're asking why did I create a `src` directory with a `psk-app` element? This is the approach the Polymer team wants to use for building apps in the future. Our CLI will use this approach and the apps the team are currently building use this structure.

index.html
robdodson: Not sure yet. Haven't thought about distribution.

src/psk-app/psk-app.html
robdodson: Yeah that's the approach we'll take. I don't think the team has produced a lazy-pages element yet (they may be working on one). For now I'll start with iron-pages

src/psk-app/psk-app.html
robdodson: Yep, I'll have to investigate.

src/psk-app/psk-app.html
robdodson: I started to go with this approach but wasn't sure if it would be bad for search engines. I'll ask the team

robdodson:
Also cc @justinfagnani and @rictic for their thoughts

src/psk-app/psk-app.html
abdonrd: Nice! :)

src/psk-app/psk-app.html
abdonrd: Okey! Waiting answer. :)

app/index.html
```diff
@@ -1,211 +0,0 @@
...
-    -->
-
-  </template>
-
-  <!-- build:js scripts/app.js -->
-  <script src="scripts/app.js"></script>
-  <!-- endbuild-->
-</body>
-
-</html>
```
addyosmani: There are a large number of files being moved to a different path here. Let's try to make sure the new structure is in line with what Polymer CLI/PolyTool will expect.

src/psk-app/psk-app.html
addyosmani: :100: Let's update all the licenses in one go.

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
...
+
+      app-drawer-layout:not([narrow]) .main-title-toolbar {
+        width: 100%;
+      }
+
+      .category-page {
+        min-height: 100vh;
+      }
+
+      @media (max-width: 580px) {
```
addyosmani: There's quite a bit of styling here that we've discussed maybe having better defaults for in the elements we're relying on. I wonder how much of this we can practically move into app-layout vs. having to override styling for ourselves.

src/psk-app/psk-app.html
addyosmani: If we're talking SEO, I think the approach you have right now might be optimal but I guess it depends on how much longer after initial render the crawler will wait to try looking at these links.

src/psk-app/psk-app.html
addyosmani: :+1: @abdonrd

src/psk-app/psk-app.html
addyosmani: Confirmed. Something weird going on here.

src/psk-app/psk-app.html
addyosmani: Thanks!

manifest.json
```diff
@@ -0,0 +1,21 @@
...
+      "type": "image/png"
+    }, {
+      "src": "//app-layout-assets.appspot.com/assets/zuperkulblog/launcher-icon-192x192.png",
+      "sizes": "192x192",
+      "type": "image/png"
+    },{
+      "src": "//app-layout-assets.appspot.com/assets/zuperkulblog/launcher-icon-384x384.png",
+      "sizes": "384x384",
+      "type": "image/png"
+    }],
```
addyosmani: Let's please make sure to use local paths for these files before landing. We should also switch `name` to `Polymer Starter Kit` so it's clear what the difference is between the name and short_name props :heart:

manifest.json
```diff
@@ -0,0 +1,21 @@
...
+  "start_url": "index.html",
```
addyosmani: If we append the add to homescreen query string to this, will it break carbon-route?

addyosmani:
Thanks for getting a start on this done @robdodson! <3 Things I would love for us to do before landing this:

- [ ] Check with @justinfagnani that we're using the application structure agreed on during design that Polytool will follow/expect
- [ ] Check if there's anything else we can drop from the new index.html to make it more DRY. Specifically was thinking about any styling pieces we can defer to the elements themselves.
- [ ] Talk about what we're doing on theming. It sounded based on our last sync with Kevin that an app-theme was fine to continue using.

addyosmani:
> I think that would be awesome if we implement an app-shell solution too.

The `app-shell` approach would require us to bake in defaults for things like skeleton layout regions by default (e.g CSS-only styles for toolbars, drawers etc). It also assumes that you'll be using the full app-shell model, including Service Worker. For the beginner, I think that may be overkill but we should talk about it.

app/index.html
robdodson: Yeah I'm trying to base the structure on the doc justin shared with us a long time ago around the directory structure he wants for CLI apps. Whatever we end up with will comply with the CLI for sure

src/psk-app/psk-app.html
robdodson: oh yeah this is all left over from zuperkulblog. I need to do an audit and clean it up. I know app-layout is trying to be as generic as possible so it may be on us to do a lot of the styling upfront so they can keep the elements vanilla

manifest.json
robdodson: Yup

manifest.json
robdodson: gooooood question

app/index.html
abdonrd: Can you share the structure?

manifest.json
abdonrd: Interesting. 😅

src/psk-app/psk-app.html
abdonrd: https://github.com/PolymerElements/carbon-route/issues/44

src/psk-app/psk-app.html
abdonrd: About this, the pesto demo (app-layout) hasn't anchors, only use the `data-route`. What is the best approach?

```html
<paper-listbox
    selected="{{_pageData.page}}"
    attr-for-selected="data-route"
    on-iron-activate="_drawerSelected">
  <paper-item data-route="art">Art</paper-item>
  <paper-item data-route="film">Film</paper-item>
  <paper-item data-route="design">Design</paper-item>
</paper-listbox>
```

And... `paper-listbox` + `paper-item` vs `iron-selector` + `a`?

manifest.json
TimvdLippe: What do you specifically mean with `append the add to homescreen query string `? I tested the query-params and they worked seemlessly with the data-bindings.

addyosmani: Applications using the Web App Manifest like airhorner.com often append a query-param string when launching the app from the homescreen to enable this to be tracked using analytics. For example: https://github.com/GoogleChrome/airhorn/blob/master/app/manifest.json#L31. I was asking if appending this might cause any issues with `carbon-route` as we have run into this with page.js in the past.

TimvdLippe: `carbon-route` will indeed detect the query-params. If the application uses any of the query-params, the object will be cluttered with the field `homescreen`. You could update this object after the page is loaded by removing this field, so I think you can work around it quite easily.

src/psk-app/psk-app.html
robdodson: I don't think a search engine will have any way of knowing it should activate those paper-item elements to navigate to the other pages so I think using anchors is a better approach.

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
...
+            <!-- menu button -->
+            <paper-icon-button
+                drawer-toggle
+                icon="app:menu">
+            </paper-icon-button>
+          </app-toolbar>
+
+          <!-- bottom toolbar -->
+          <app-toolbar class="title-toolbar main-title-toolbar">
+            <div class="title">STARTER KIT</div>
```
chuckh: Shouldn't this be Polymer Starter Kit?  If so `.title` font size will need to be `50px` instead of `60px`.

```
.title {
     padding-bottom: 40px;
     font-size: 50px;
     font-weight: 800;
 }
```

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
...
+
+      .title-toolbar {
+        @apply(--layout-center-center);
+        width: 100vw;
+        box-sizing: border-box;
+      }
+
+      .title {
+        padding-bottom: 40px;
+        font-size: 60px;
```
chuckh: Consider changing to `50px` to better handle long titles such as Polymer Starter Kit.

src/psk-app/psk-app.html
chuckh: Consider using generic names for pages. This way PSK user does not have to rename.

```
<page-one data-page="art"></page-one>
<page-two data-page="film"></page-two>
<page-three data-page="design"></page-three>
```

TimvdLippe: @chuckh This will incentive Polymer developers to use non-descriptive names for their own elements. Therefore I think it is better to use more descriptive names such as `psk-page-art` as this is a lot more readible. Whenever you are downloading the PSK, you will delete quite some code that is specific to the PSK anyways. It is primarily meant as example and skeleton.

src/psk-app/psk-app.html
```diff
@@ -0,0 +1,226 @@
...
+
+    </app-drawer-layout>
+
+  </template>
+
+  <script>
+
+    Polymer({
+
+      is: 'psk-app',
```
chuckh: Consider using more generic name for `psk-app` such as `poly-app` or `app-start` or `app-begin` or `app-main`, etc.  This way PSK user does not have to rename.

src/psk-app/psk-app.html
abdonrd: I'm agree with @TimvdLippe. :)

chuckh:
Thanks @robdodson for getting this started! :smile: :100: Some additional items we should consider doing before landing:

- [ ] Discus how best revise/organize README.md
- [ ] Add the 3 internal pages
- [ ] Add basic tests or instructions how to do so
- [ ] Add instructions on how to adding tooling (gulp, bowsersync, deploying to github, etc, similar to what PSK1 has)
- [ ] Add instructions on how to use material design colors
- [ ] Add back PSK icons or instructions how to do so

I think it is important for us to balance simplicity for beginning users why still making PSK useful for immediate and advanced users.

szeck87:
Good news this is starting! Is there a defined MVP for v2 besides the high level goals described in https://github.com/PolymerElements/polymer-starter-kit/issues/653?

app/index.html
justinfagnani: I'll do this in a couple of days when I'm back from vacation: https://github.com/Polymer/polytool/issues/13

Meanwhile, you can see the project layouts in https://github.com/PolymerLabs/generator-polymer-init

src/psk-app/psk-app.html
justinfagnani: I think hardcoding is fine in index.html - it's the document after all. I can see the desire to have the menu possibly driven by a CMS however, so I wonder how such variants could be expressed in PSK? Maybe a recipie/tutorial that replaces this with a template or custom menu element?

app/index.html
abdonrd: Thanks @justinfagnani! :grin:

src/psk-app/psk-app.html
rictic: Hm, two things I notice.

* Try removing the # and just doing `this.set('route.path', '/art');`
* It's best to do this in response to route.path changing, or at least inside of `attached`, because before then you may not be able to be confident that route.path has been initialized from the URL. `iron-location` ignores all writes to itself before its attached callback fires so that it properly respects the initial page URL.

manifest.json
rictic: Yep. Query params won't break anything with carbon-route, it will correctly handle URLs with paths, hashes, query params or any combination of the three.

As @TimvdLippe says, the query params will be exposed to the data-binding system as well, in case that's useful.

src/psk-app/psk-app.html
robdodson: > It's best to do this in response to route.path changing, or at least inside of attached, because before then you may not be able to be confident that route.path has been initialized from the URL. iron-location ignores all writes to itself before its attached callback fires so that it properly respects the initial page URL.

Good point. I stole that code from https://github.com/PolymerLabs/app-layout/blob/master/templates/publishing/src/blog-app.html#L302-L308 so we may want to update our app-layout templates as well

abdonrd: I tried with:

##### Option 1: working for me

```javascript
attached: function() {
  if (!this.route.path) {
    this.set('route.path', '/art');
  }
},
```

##### Option 2: working for me

```javascript
observers: [
  '_routePathChange(_route.path)'
],

_routePathChange: function() {
  this.async(function() {
    if (!this.route.path) {
      this.set('route.path', '/art');
    }
  });
},
```

But... why observe the path changes? We only want to set the default value the first time. Right?

app/index.html
chuckh: Below is the current application structure that `polytool init` creates:

image

manifest.json
```diff
@@ -0,0 +1,21 @@
...
+  "background_color": "#fafafa",
+  "display": "standalone",
+  "theme_color": "#fafafa"
```
kenchris: wouldn't it be better to use different colors for theme and background?

src/psk-app/psk-app.html
kenchris: I did an attempt of lazy loading in my project:

https://github.com/01org/webnfc-shoppingcart-demo/blob/master/app/elements/main-app/main-app.html#L395
https://github.com/01org/webnfc-shoppingcart-demo/blob/master/app/scripts/utils.js#L22

motss:
Is it open to contributions? Perhaps do something like `app-*` elements, a repo that all of us can try/ experiment on the next version of `PSK`.

samccone:
Hey all, we have the psk2 branch now, I am going to close this PR and please open pull requests into that branch to address your individual comments on this thread!

abdonrd:
I has created a [v2.0.0 label](https://github.com/PolymerElements/polymer-starter-kit/labels/v2.0.0). :)

abdonrd:
@robdodson @justinfagnani with the new `polytool` structure, where we add `.css` files? `src/styles/*.css`?

robdodson:
@abdonrd I think the idea is they would live with the elements. src/psk-app/psk-app.css

abdonrd:
@robdodson I was referring to some generic styles like [PSK1 app/styles/](https://github.com/PolymerElements/polymer-starter-kit/tree/master/app/styles)

robdodson:
@abdonrd I know but I don't think we'll have those anymore as the entire app will be contained in an element :)

abdonrd:
@robdodson but what happens with styles like this?

```css
body {
  margin: 0
}
```

robdodson:
Good question, @justinfagnani wdyt?

justinfagnani:
they can go in index.html or index.css?

PR #771, opened 2016-04-06
New "Add push notifications with platinum-push-messaging recipe"

nagarjunredla:
Added a recipe to get started with push messaging on PSK.


addyosmani:
cc @kaycebasques @arthurevans

petrc:
When running these changes with `gulp serve:dist` the platinum-push-message service-worker.js cannot be found (404) as it tries to load form the elements directory

arthurevans:
@kaycebasques can you take a look at this one?

robdodson:
Thanks for the PR @nagarjunredla and sorry it's taken so long to follow up.

We've recently changed the structure of PSK and we're no longer relying on the platinum elements. I think this is the kind of thing that may make more sense as a blog post on something like [Medium](medium.com).

PR #776, opened 2016-04-10
Add missing paper-styles/color.html

abdonrd:
Add missing paper-styles/color.html, used in :host --paper-indigo-500


robdodson:
Thanks!

PR #777, opened 2016-04-10
Update bower.json and package.json

abdonrd:


bower.json
```diff
@@ -1,16 +1,12 @@
...
-    "polymer": "Polymer/polymer#^1.0.0",
-    "carbon-route": "PolymerElements/carbon-route#~0.8.4",
-    "app-layout": "PolymerLabs/app-layout#~0.0.14"
-  },
-  "devDependencies": {
-    "web-component-tester": "^4.0.0"
+    "polymer": "Polymer/polymer#^1.0.0"
   },
-  "ignore": []
+  "private": true
```
robdodson: why move this?

bower.json
```diff
@@ -1,16 +1,12 @@
 {
   "name": "polymer-starter-kit",
-  "private": true,
   "dependencies": {
+    "app-layout": "PolymerLabs/app-layout#~0.0.14",
```
robdodson: I think we should switch these to use `^` style

package.json
```diff
@@ -1,9 +1,19 @@
...
+    "type": "git",
+    "url": "git@github.com:PolymerElements/polymer-starter-kit.git"
+  },
+  "license": "http://polymer.github.io/LICENSE.txt",
+  "bugs": {
+    "url": "https://github.com/PolymerElements/polymer-starter-kit/issues"
+  },
   "devDependencies": {
-
+    "web-component-tester": "^4.2.0"
```
robdodson: I'm wondering if this should be optional or not. It's a huge dependency to download if I recall

bower.json
abdonrd: Nothing special. To have it in the same place in both files: `bower.json` and `package.json`.
And because it's something that will not be changes, have it at the end of the file.

bower.json
abdonrd: Right!

package.json
abdonrd: I would prefer it to be optional.
I will remove for now.

abdonrd:
Update with the comments!

bower.json
```diff
@@ -1,16 +1,12 @@
 {
   "name": "polymer-starter-kit",
-  "private": true,
   "dependencies": {
+    "app-layout": "PolymerLabs/app-layout#~^0.0.15",
+    "carbon-route": "PolymerElements/carbon-route#~^0.8.4",
```
chuckh: Did you mean this to be `carbon-route#^0.8.4""`  and app-layout#^0.0.15", removing the '~'?

abdonrd: My fault! Really fixed now. Thanks @chuckh! :grin:

robdodson:
LGTM thanks!

PR #778, opened 2016-04-10
Clean project

abdonrd:
I propose remove this unused files, since now the structure is different.
We may be creating as we need.


robdodson:
Hm that's a question for @samccone and @chuckh. I'm not super familiar with how the travis stuff works

abdonrd:
@robdodson okey!

If we don't write the:

```yml
script:		
  - ./travis-runner.sh
```

Travis will be run `npm test` by default.

And because the structure has changed, the content of `travis-runner.sh` not working. We must to update the it.

chuckh:
I think we should leave them for now as place holders. We will still need them to handle auto deploying to Github and Firebase. Once we get closer to agreeing on the new structure they can be updated.

abdonrd:
While we are not sure how it will be, I think it would be better to have it clean.
We always have the git history and the `master` branch with the previous changes.

abdonrd:
I would also add `gulp lint` in `.travis.yml`:

```yml
before_script:
 ...
  - npm i -g bower gulp
  - bower i
  - gulp lint
```

addyosmani:
> I think we should leave them for now as place holders. We will still need them to handle auto deploying to Github and Firebase. Once we get closer to agreeing on the new structure they can be updated.

I agree with Chuck on this one.

abdonrd:
Then we close it?

chuckh:
Yes. Closing it now.

PR #779, opened 2016-04-10
Add ESLint support

abdonrd:


robdodson:
I think we're still trying to figure out what we want our gulp support to look like. What we want to avoid is the crazy gulpfile that is currently in PSK1. Our current thinking is to break each gulp task into its own module and use a similar approach to the one used in wsk2. https://github.com/google/web-starter-kit/tree/wsk2/src

We could see about doing something like that with this eslint task maybe?

abdonrd:
Makes sense! I will update. :)

abdonrd:
@robdodson updated! :sparkles:

chuckh:
:+1:

robdodson:
I'm wondering if it would be possible to use the same gulpfile approach as WSK does here https://github.com/google/web-starter-kit/blob/wsk2/src/gulpfile.js

Using require-dir makes all of the gulp tasks mandatory but I think we may want to make it clear which ones are being used, and use the same configuration approach as WSK. That way users can have a bit more control over the behavior of the gulpfile

abdonrd:
@robdodson sounds good to me. And we avoid one more npm dependency. :) Check it now.

I code:

```javascript
require('./gulp-tasks/lint.js');
```

Instead of:

```javascript
var lint = require('./gulp-tasks/lint.js');
```

To avoid the `no-unused-vars` error, because nobody will use the `lint` var.

abdonrd:
@robdodson updated and rebased! :)

chuckh:
LGTM. Nice work @abdonrd!

robdodson:
Was off filming stuff today but I'll try to review this in the morning. Thanks for updating the PR @abdonrd :D

abdonrd:
Your welcome! :)

ESLint support will be nice in PSK2! :grin:

abdonrd:
Friendly ping to @robdodson. :)

robdodson:
@justinfagnani how would you like us to proceed on this?

justinfagnani:
Can we add JS linting to `polytool lint`? Contributions welcome :)

robdodson:
@justinfagnani we can I guess.. how should we decide which files to source? At present they're done like this in the gulpfile https://github.com/PolymerElements/polymer-starter-kit/pull/779/files#diff-742af4f7ce30ade9ca8aab1f4046729cR18

abdonrd:
Now I can't contribute much in the polylint, because I don't how it works. 😅

abdonrd:
But then... `polytool` will manage all the linting? `ESLint` + `PolyLint`?

robdodson:
@abdonrd I guess that's the plan. @justinfagnani there may be some thing which don't make sense as polytool commands, do you have an example of a project gulpfile that gets run as part of polytool?

addyosmani:
One thing we probably want to sanity check if we do add ESLint support to `polytool` is that ESLint configuration is still low-friction for folks and picks up rules from `.eslintrc`/`package.json` etc. Anywhere they're likely to try configuring these pieces.

justinfagnani:
Let's chat about linting this week. Maybe JS linting doesn't belong in polylint? Polytool can certainly get the hook for retrieving all the files it cares about, for use in a gulfile say. This includes splitting script from HTML if necessary.

abdonrd:
@justinfagnani yes, I think that `polylint` only lint Polymer. But `polytool lint`, could have `polylint`plus another configurable (by gulp tasks?) linters (like `eslint`).

abdonrd:
@justinfagnani any news about how to opt-in the `eslint` to the Polymer CLI lint (`polymer lint`)? Thanks in advance!

abdonrd:
PR updated! @robdodson @chuckh PTAL :)

robdodson:
@justinfagnani do you think this is how you'd like something like ESLint used or would you prefer to see it go through the splitHTML pipeline simile to the way we do imagemin?

abdonrd:
Friendly ping to @justinfagnani. :)

abdonrd:
@robdodson @chuckh @justinfagnani updated with the latest PR (https://github.com/PolymerElements/polymer-starter-kit/pull/886): Update to polymer-build v0.3.0

abdonrd:
After thinking about that, maybe we need to update the tasks with `module.exports` as the [google/web-starter-kit](https://github.com/google/web-starter-kit/tree/wsk2/src)?

abdonrd:
Closed in favor to [PolymerElements/generator-polymer-init-custom-build](https://github.com/PolymerElements/generator-polymer-init-custom-build).

PR #780, opened 2016-04-10
Add pages

abdonrd:


src/psk-app/psk-app.html
```diff
@@ -15,6 +15,7 @@
 
 <link rel="import" href="../../../iron-flex-layout/iron-flex-layout.html">
 <link rel="import" href="../../../iron-pages/iron-pages.html">
+<link rel="import" href="../../../iron-pages/iron-pages.html">
```
robdodson: you're importing this twice

robdodson:
This seems like a good start. We may end up renaming the app from `psk-*` to just like `my-*`, I'm not sure but i think we can go ahead and merge this and deal with renaming stuff later if we decide to do that.

src/psk-app/psk-app.html
abdonrd: Oops! I will remove.

abdonrd: Fixed here: https://github.com/PolymerElements/polymer-starter-kit/commit/47007581395eeaadcb17f3b08ae75e8be5d242d8

PR #782, opened 2016-04-10
PSK2 update .gitignore for deploy

chuckh:
Until we decide not to use `gulp` with PSK2 we need to ignore `dist`.
Until we change our method to deploy to Github Pages we will still want
`.tmp` and `.publish/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 bower_components/
 node_modules/
+dist
+.tmp
+.publish/


abdonrd:
Right now we haven't a build task in the gulpfile.
I prefer update the `.gitignore` when we create the build task, not before. :confused:

chuckh:
By not having these ignored may cause others issues who work both on PSK1 and PSK2. By having these now will avoid that.

robdodson:
I see. I think @chuckh is concerned that if he's working on PSK1 and then switches to PSK2 branch that it may accidentally commit one of the above mentioned directories. I'm ok adding this in for now if only to avoid us accidentally committing a bunch of stuff. We can circle back and clean the ignore file once things are more settled.

PR #783, opened 2016-04-10
Update manifest.json

abdonrd:
Update `manifest.json` with previous PSK colors, `/?homescreen=1` url and all the icons


robdodson:
👍

PR #786, opened 2016-04-11
Add 404 section to iron-pages with fallback-selection

TimvdLippe:
This PR adds a 404 default message when a specific page could not be found. It uses the recently added property `fallback-selection` (ref: https://github.com/PolymerElements/iron-selector/pull/117)

### How to test?

1. Open the PSK in your browser
* Modify the url by replacing `art` with a non-existing page (for example `asdf`)
* Observe the 404 message is displayed and the URL has been changed to `/#/404`

The last step is due to the two-way binding on the `selected` property. If you do not want to modify the url (e.g. keep the url at `/asdf`) I can revert back to one-way binding.

### DIFF
--- a/src/psk-app/psk-app.html
+++ b/src/psk-app/psk-app.html
@@ -77,6 +77,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         background-color: #888;
       }
 
+      iron-pages a {
+        color: var(--app-primary-color);
+      }
+
       .main-header {
         border-bottom: 1px solid #ddd;
         background-color: #fff;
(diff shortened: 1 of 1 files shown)


chuckh:
LGTM  :smile:

src/psk-app/psk-app.html
```diff
@@ -173,10 +177,15 @@ <h4 class="title">Polymer Starter Kit - [[params.page]]</h4>
         </app-header>
       </app-header-layout>
 
-      <iron-pages selected="[[params.page]]" attr-for-selected="data-page">
+      <iron-pages selected="{{params.page}}"
```
robdodson: any reasons why this should switch to a two way binding?

robdodson:
LGTM, just have the one comment

src/psk-app/psk-app.html
TimvdLippe: As noted in the description, this will modify the url to `/404`.

abdonrd:
Nice work!

Personally, I prefer:
* One-way binding for no change the URL.
* Keep the iron-pages in one line. It isn't very large.

src/psk-app/psk-app.html
robdodson: ah i see :)

robdodson:
ack! sorry @abdonrd I didn't see your comment

abdonrd:
Haha no problem, @robdodson. :smile: 

@robdodson @chuckh @TimvdLippe what do you think about [my comments](https://github.com/PolymerElements/polymer-starter-kit/pull/786#issuecomment-208481801)?

chuckh:
@abdonrd I agree with your comments.

robdodson:
I think I'd prefer if it left the wrong url in the bar and just displayed the 404 page so I think that would mean switching this to one-way, yeah?

TimvdLippe:
Fine by me. Then it should be `selected=[[params.page]]`

abdonrd:
Nice! I will create a PR!

PR #787, opened 2016-04-11
Update 404 page

abdonrd:
Reference: https://github.com/PolymerElements/polymer-starter-kit/pull/786#issuecomment-208481801

### DIFF
--- a/src/psk-app/psk-app.html
+++ b/src/psk-app/psk-app.html
@@ -31,6 +31,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../../../app-layout/app-toolbar/app-toolbar.html">
 
 <link rel="import" href="../app-icons/app-icons.html">
+<link rel="import" href="../psk-pages/psk-page-404.html">
 <link rel="import" href="../psk-pages/psk-page-art.html">
 <link rel="import" href="../psk-pages/psk-page-design.html">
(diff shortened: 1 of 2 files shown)


robdodson:
LGTM thanks!

chuckh:
:+1:  

Also it would be nice to have a home page `http://localhost:8080/components/polymer-starter-kit/#/` instead routing to `http://localhost:8080/components/polymer-starter-kit/#/art`

abdonrd:
@chuckh yes! Can you send a PR? :smiley:

chuckh:
@abdonrd I will give it a try now.

robdodson:
Hmm that means we need to come up with a design for that page. I'd prefer to stick as close to the current zuperkulblog as possible

chuckh:
@robdodson I understand, but I think having a home page is a common use case like we have in PSK1 and is worth the effort.

robdodson:
Before we spend too much time on that I think it'd be good to get everything else to a near final point and make sure that things are playing nice with the CLI and everyone is happy with the basic shape of the site. In the back of my mind I'm wondering if we're going to end up scrapping stuff as requirements change. We may want PSK to just turn into 1 of many simplified templates that all live in a single repo

robdodson:
Sorry to continue on from that thought a bit. I'm cool with the idea of adding a home page but don't want folks to be annoyed if we end up scrapping that work because we decide we want the template to be something simpler.

PR #788, opened 2016-04-11
Rename route and add to properties

abdonrd:


robdodson:
Hm I'm kind of on the fence about these changes just from a pure aesthetic perspective. I prefer the look of things in the previous version. I guess route and params are private but don't know if that is a huge deal because they're at the app level so I liked it better without the underscores.

Similarly, I prefer params to routeData only because I think the word data is a "noise word." It doesn't actually describe what it is. The book Clean Code talks about this and it's always stuck with me. "data" is like saying "stuff", it doesn't mean anything to your audience. I guess I went with params originally to have some continuity with Express and page.js style routers.

abdonrd:
I was inspired by [Pesto](https://github.com/PolymerLabs/app-layout/blob/master/templates/pesto/src/recipe-app.html#L168) (app-layout demo).
But okey! I will remove the underscores.

About the params... I update it to `_dataRoute` because the `carbon-route` attribute is called `data`.

Then. Close this PR and create one that only add the actual variables to attributes?

robdodson:
I'm not even sure if we need the variables in the properties object.

abdonrd:
@robdodson okey! Can you check with the team which is the best practice? :)

src/psk-app/psk-app.html
```diff
@@ -191,9 +191,19 @@ <h4 class="title">Polymer Starter Kit - [[params.page]]</h4>
     Polymer({
       is: 'psk-app',
 
+      properties: {
+        _route: {
+          type: Object
+        },
+
+        _routeData: {
```
samccone: since this is not private and is used in the templates, using _ seems misleading ... thoughts?

abdonrd: Yes! Check: https://github.com/PolymerElements/polymer-starter-kit/pull/788#issuecomment-208674908 and https://github.com/PolymerElements/polymer-starter-kit/pull/788#issuecomment-208855399. :)

abdonrd:
@robdodson have you news about https://github.com/PolymerElements/polymer-starter-kit/pull/788#issuecomment-208964666? :)

robdodson:
Just based on looking at other carbon-route projects I decided to leave `route` and rename `params` to `pageData` which matches the `/:page` token

robdodson:
Added in 605d8fc

abdonrd:
@robdodson okey! But I meant about https://github.com/PolymerElements/polymer-starter-kit/pull/788#issuecomment-208951962 and https://github.com/PolymerElements/polymer-starter-kit/pull/788#issuecomment-208964666; We should add the properties?

robdodson:
@abdonrd I think we can leave the properties out. They don't serve much of a purpose in my opinion since we're not trying to expose them as a public API

abdonrd:
@robdodson perfect! 👌

PR #789, opened 2016-04-11
Lazy load polyfill if browser does not support native webcomponents

TimvdLippe:
Based on https://gist.github.com/ebidel/1d5ede1e35b6f426a2a7

@justinfagnani confirmed that this should be compatible with the CLI as it just uses `polyserve`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,12 +18,39 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <meta name="theme-color" content="#fafafa">
 
     <link rel="manifest" href="manifest.json">
-
-    <script src="../webcomponentsjs/webcomponents-lite.js"></script>
-
     <link rel="import" href="src/psk-app/psk-app.html">
   </head>
   <body>
     <psk-app></psk-app>
+
+    <script>
+      // Load webcomponentsjs polyfill if browser does not support native webcomponents
+      (function() {
+        'use strict';
+        var onload = function() {
+          // For native Imports, manually fire WCR so user code
+          // can use the same code path for native and polyfill'd imports.
+          if (!window.HTMLImports) {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -19,7 +19,35 @@
 
     <link rel="manifest" href="manifest.json">
 
-    <script src="../webcomponentsjs/webcomponents-lite.js"></script>
+    <script>
```
TimvdLippe: Fixed :+1:

abdonrd:
Nice work, @TimvdLippe! :tada: 

My unique little recommendation (and opinionated) is indent the script inside the `script` tags.

TimvdLippe:
Just like the original gist is not indented, I personally prefer to not indent the code.

robdodson:
I guess we need to come up with a style convention. I also tend to indent in the script block :P

What're we doing for script tags in PSK1?

TimvdLippe:
Can ESlint enforce this style convention? It seems that PSK1 indeed indents the code. I will update the PR

abdonrd:
@robdodson totally! Style guides. :)

@TimvdLippe yes! with this: https://github.com/abdonrd/polymer-starter-kit/blob/eslint_support/.eslintrc.json#L21-L22

abdonrd:
@TimvdLippe rebase on `psk2` branch and I think that we can merge! :)

TimvdLippe:
Rebased :)

abdonrd:
@TimvdLippe sorry, the script indent. 😅

TimvdLippe:
Argh I fixed that online, but the rebase overwrote it >.<

abdonrd:
Thanks @TimvdLippe! 🎉

PR #775, opened 2016-04-10
Enabling caching of node_modules on Travis

abdonrd:
Reference: https://github.com/Polymer/polymer/pull/3510

PR #793, opened 2016-04-14
Revert "Enabling caching of node_modules on Travis"

samccone:
Reverts PolymerElements/polymer-starter-kit#775

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: node_js
 sudo: false
-node_js:
-  - '0.12'
-  - '4.2'
-  - '5.1'
 addons:
   firefox: latest
   apt:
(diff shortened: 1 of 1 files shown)


TimvdLippe:
What was the reason to revert this change?

samccone:
it was causing CI to fail.

abdonrd:
We know why it failed? The https://github.com/PolymerElements/polymer-starter-kit/pull/795 fixes?

TimvdLippe:
I think indeed the failures are fixed in #795, as the builds show that `apt-get install` actually failed (and not `npm install`) https://travis-ci.org/PolymerElements/polymer-starter-kit/jobs/122548438

PR #795, opened 2016-04-14
Migrate to Travis trusty

### LINKED ISSUE #794: Migrate to Travis Trusty images for Chrome testing
Chrome 50 drops Ubuntu Precise, which is the default Travis CI base image.
The only workaround is to use the Trusty images by adding
```yaml
sudo: required
dist: trusty
```
to the travis config

azakus:
Trusty has gcc 4.8, drop ubuntu-toolchain-r-test ppa

Fixes #794

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,13 @@
 language: node_js
-sudo: false
+sudo: required
+dist: trusty
 addons:
   firefox: latest
   apt:
     sources:
       - google-chrome
-      - ubuntu-toolchain-r-test
     packages:
       - google-chrome-stable
-      - g++-4.8
 node_js:
   - '0.12'
   - '4.2'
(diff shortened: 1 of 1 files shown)


samccone:
There are other options here, we could download chrome directly and cache it so we do not loose the build time optimizations (because of sudo)

Here is how I handle it:
https://github.com/GoogleChrome/lighthouse/blob/master/test/download-chrome.sh
https://github.com/GoogleChrome/lighthouse/blob/master/.travis.yml#L16

addyosmani:
Sam: are there any known downsides to your aporoach? I'd like to get consistent with how the rest of Polymer is tackling this but if there are build time optimisations we get to keep using your way I'm in favor of that.

samccone:
sure let's be consistent and see how much slower this gets, we can always tweak this.

PR #798, opened 2016-04-18
Migrate PSK2 to Travis trusty

chuckh:
Trusty has gcc 4.8, drop ubuntu-toolchain-r-test ppa.  This matches
PSK1 master. Based on @azakus changes to PSK1.


abdonrd:
Thanks! 😁

PR #799, opened 2016-04-19
Sample content for all pages and style cleanup

robdodson:
screen shot 2016-04-18 at 9 49 56 pm

- Add `sample-content` taken from app-layout examples
- Switch to greyscale colors
- Remove unused styles
- Add default `lang` to make a11y audit tools happy

### DIFF
--- a/src/psk-app/psk-app.html
+++ b/src/psk-app/psk-app.html
@@ -41,9 +41,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   <template>
     <style>
       :host {
-        --app-primary-color:  var(--paper-indigo-500);
+        --app-primary-color:  var(--paper-grey-800);
         -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+
+        display: block;
         font-family: 'Roboto', sans-serif;
+        background-color: var(--paper-grey-50);
       }
 
       app-drawer {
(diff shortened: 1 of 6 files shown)


abdonrd:
Instead create the `sample-content`... Why not import the `app-layout/demo/sample-content.html`?

robdodson:
@abdonrd good call!

robdodson:
@abdonrd Rebased, PTAL

src/psk-app/psk-app.html
```diff
@@ -76,13 +79,17 @@
...
-        background-color: var(--paper-indigo-800);
+        background-color: var(--paper-grey-600);
       }
 
       iron-pages a {
         color: var(--app-primary-color);
       }
 
+      iron-pages > * {
+        min-height: 100vh;
```
abdonrd: We need this?

abdonrd:
Testing this PR, I see that the page need something like:

```css
body {
  margin: 0;
}
```

src/psk-app/psk-app.html
robdodson: I believe so. If you remove the sample-content the page collapses. I figure there will be folks who put in content that doesn't 100% fill up the vertical height of the screen and this will prevent it from looking weird for them.

robdodson:
@abdonrd updated, PTAL

src/psk-app/psk-app.html
abdonrd: 👍

abdonrd:
Yay! Now we have content! 🎉

robdodson:
\o/

robdodson:
Thanks for the review @abdonrd 👍

abdonrd:
Thanks for the work @robdodson! 😄

PR #801, opened 2016-04-19
Add Page Titles

chuckh:
Add page title to each page. When using PSK2 on mobile you can
determine which page you are on without looking at url or opening menu.
 Also made Design page content size 2 to have one short page for
testing.

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,6 +24,20 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       body {
         margin: 0;
       }
+
+      /* shared style for page titles */
+      .page-title {
+        font-size: 40px;
+        margin-left: 40px;
+        color: #37474F;
+      }
+
+      @media (max-width: 580px) {
+        /* make page title smaller */
+        .page-title {
+          font-size: 26px;
+        }
+      }
     </style>
 
     <link rel="import" href="src/psk-app/psk-app.html">
--- a/index.html
+++ b/index.html
@@ -24,20 +24,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       body {
         margin: 0;
       }
-
-      /* shared style for page titles */
-      .page-title {
-        font-size: 40px;
-        margin-left: 40px;
-        color: #37474F;
-      }
-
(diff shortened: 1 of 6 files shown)


src/psk-page-design/psk-page-design.html
```diff
@@ -19,7 +19,8 @@
       }
     </style>
 
-    <sample-content size="100"></sample-content>
+    <h2 class="page-title">Design</h2>
+    <sample-content size="2"></sample-content>
```
robdodson: Any reason you changed this one to 2 instead of 100?

chuckh: Yes to demonstrate how a page that is shorter then the normal full height of a page will look and operate.

index.html
```diff
@@ -24,6 +24,20 @@
       body {
         margin: 0;
       }
+
+      /* shared style for page titles */
+      .page-title {
+        font-size: 40px;
+        margin-left: 40px;
+        color: #37474F;
```
robdodson: Do we have a color var we could use here?

robdodson: oh wait we wouldn't want this style here. I think it needs to be part of a shared stylesheet that the pages themselves use...

chuckh: That color is paper-grey-800, but this variable is not available in index.html.

I put it in index.html since we do not have a shared stylesheet.  I can give it a try adding a shared stylesheet.

abdonrd: +1 to shared *style module*.

chuckh:
Added shared styles.

src/psk-page-design/psk-page-design.html
chuckh: Also when testing with design page noticed that scroll position is incorrect when one has scrolled down on another page.  We probably will want to do something to scroll to top when changing pages like PSK1 does.

robdodson:
Cool thanks Chuck. I'm following up with Justin on Slack to see if this is inline with his directory structure plan

addyosmani:
LGTM pending Rob checking up with Justin on structure.

robdodson:
LGTM. Justin ok'd the structure

PR #802, opened 2016-04-19
Add Scroll Position Control

chuckh:
app-scrollpos-control is a manager for saving and restoring the scroll position when multiple pages are sharing the same document scroller.


abdonrd:
LGTM! :)

addyosmani:
LGTM 2.

PR #810, opened 2016-04-29
Update bower_components path to support polytool serve

abdonrd:
Update `bower_components` path to support `polytool serve`.

The new `generator-polymer-init` already use these paths: https://github.com/PolymerLabs/generator-polymer-init/blob/master/generators/app/templates/application/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -26,7 +26,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
     </style>
 
-    <link rel="import" href="src/psk-app/psk-app.html">
+    <link rel="import" href="/src/psk-app/psk-app.html">
   </head>
   <body>
     <psk-app></psk-app>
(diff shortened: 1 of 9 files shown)


abdonrd:
@justinfagnani it is correct, right? :)

index.html
```diff
@@ -52,7 +52,7 @@
         if (!webComponentsSupported) {
           var script = document.createElement('script');
           script.async = true;
-          script.src = '../webcomponentsjs/webcomponents-lite.min.js';
+          script.src = 'bower_components/webcomponentsjs/webcomponents-lite.min.js';
```
justinfagnani: This should be an absolute URL if index.html is loaded from any other path, as in shell apps that serve index from `/foo/bar`.

justinfagnani: The HTML import for psk-app should be absolute too

justinfagnani:
LGTM

abdonrd:
Update with Justin comments!

addyosmani:
👍

PR #814, opened 2016-05-03
Change to pretty urls

chuckh:
Now that latest `polyserve` / `polytool serve` support pretty urls let’s
make it the default approach.

### DIFF
--- a/src/psk-app/psk-app.html
+++ b/src/psk-app/psk-app.html
@@ -130,7 +130,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     </style>
 
     <!-- setup routes -->
-    <carbon-location route="{{route}}" use-hash-as-path></carbon-location>
+    <carbon-location route="{{route}}"></carbon-location>
     <carbon-route route="{{route}}" pattern="/:page" data="{{pageData}}"></carbon-route>
 
     <app-scrollpos-control selected="{{pageData.page}}"></app-scrollpos-control>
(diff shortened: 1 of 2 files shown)


abdonrd:
LGTM!

But... I think that we need to avoid hardcoded the URLs.

robdodson:
@abdonrd did you have something else in mind for the URLs?

chuckh:
Per @arthurevans on Slack 

> Using `<a>` tags for navigation is important for SEO. It lets crawlers find the links. It also enables you to do webby things like right-click/command-click to open in another tab.

Lets stay with hardcode URLs such as: `<a data-page="art" href="#/art">Art</a>` for now.

robdodson:
I'm ok with hard coded URLs especially since we want to keep the template quite simple.

abdonrd:
@chuckh Sure! SEO is important.

@robdodson About the hardcoded URLs... Maybe something like this?

```html
<iron-selector
    class="nav-menu"
    selected="[[pageData.page]]"
    attr-for-selected="data-page"
    on-iron-activate="_drawerSelected">
  <a data-page="" href="{{routes.home}}">Home</a>
  <a data-page="art" href="{{routes.art}}">Art</a>
  <a data-page="film" href="{{routes.film}}">Film</a>
  <a data-page="design" href="{{routes.design}}">Design</a>
</iron-selector>
```

```html
Oops you hit a 404. <a href="{{routes.home}}">Head back to home.</a>
```

```javascript
properties: {
  routes: {
    type: Object,
    value: function() {
      return {
        home: '/',
        art: '/art',
        film: '/film',
        design: '/design'
      }
    }
  }
}
```

robdodson:
@abdonrd I think for keeping the code size small we can stick with hard coded urls. I can see the value of having the bindings as it lets you share and reuse the urls as named routes, but I'm also a little unsure if that means we would need to duplicate that object in multiple parts of the app. Kinda curious what the carbon-route folks think. /cc @rictic

PR #821, opened 2016-05-07
PSK2 Added .bowerrc file so that bower components end up in the right foder

mercmobily:
At the moment, running a bower_install will install everything in a freshly created "app" directory, which won't be found by the component.
With this file, bower components will end up where they are meant to go

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory" : "bower_components"
+}


abdonrd:
I haven't this problem.
Can not be... 😕 If we haven't `.bowerrc`, the default path is `bower_components`.

jsilvermist:
Sounds to me like a leftover file from psk1 got in your psk2 when you did `bower install`.

robdodson:
As mentioned by @jsilvermist, make sure you don't have an old .bowerrc from PSK1 laying around. You shouldn't need a .bowerrc with PSK as the default location (when no .bowerrc is present) is to install to `bower_components.

mercmobily:
I am sorry -- you guys were right. I recloned https://github.com/PolymerElements/polymer-starter-kit.git and checked out psk2 and .bowerrc isn't there. I really have no idea why it was left over in the first place -- and why gig status didn't tell me about it. Weird... but sorry about the noise.

PR #822, opened 2016-05-07
PSK2 Swipe effect for drawer and condense effect for header

mercmobily:
With these simple modifications, the basic PSK2 app will get the nice effects enjoyed by PS1 -- the swipe-open drawer and the condensing header. When I installed it, I wouldn't work out why the app looked so "stiff"...

### DIFF
--- a/src/psk-app/psk-app.html
+++ b/src/psk-app/psk-app.html
@@ -103,7 +103,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
 
       .title {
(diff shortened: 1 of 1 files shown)


abdonrd:
Thanks @mercmobily!
I propose another option at https://github.com/PolymerElements/polymer-starter-kit/pull/825

mercmobily:
#825 is a billion times better.

abdonrd:
Thanks! 😄

PR #823, opened 2016-05-07
Separate theming file for app, in line with other page files and easier to deal with/change

mercmobily:
It really bothered me that PSK2's theming options were inline within psk-app.html. I like the fact that we no longer use a "general" CSS that gets included, and the fact that CSS is set at element-level.
I think it makes a lot more sense this way. Plus, the app looks better too...

### DIFF
--- a/src/psk-app/psk-app.html
+++ b/src/psk-app/psk-app.html
@@ -36,98 +36,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../psk-page-art/psk-page-art.html">
 <link rel="import" href="../psk-page-design/psk-page-design.html">
 <link rel="import" href="../psk-page-film/psk-page-film.html">
+<link rel="import" href="../styles/app-theme.html">
 
 <dom-module id="psk-app">
   <template>
-    <style>
-      :host {
-        --app-primary-color:  var(--paper-grey-800);
-        -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
-
-        display: block;
-        font-family: 'Roboto', sans-serif;
-        background-color: var(--paper-grey-50);
-      }
-
-      app-drawer {
-        --app-drawer-content-container: {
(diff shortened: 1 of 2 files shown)


abdonrd:
I'm not sure if it is the best option, but I prefer a theme file with the colors and little else, not all the CSS.

mercmobily:
So, the app-theme.html could be something like this, ONLY to include the general app-wise CSS custom properties (and leaving the CSS belonging to the `psk-app.html` where it is now..._ 

The rationale is that changes here will affect the WHOLE app, whereas changes to psk-app.html will only affect the "app" element as such (which makes sense).

    
    <link rel="import" href="../../bower_components/polymer/polymer.html">
    <link rel="import" href="../../bower_components/paper-styles/color.html">

    
    <dom-module id="app-theme">
      <template>
        <style>
          :host {

            /* PSK2-specific custom properties */
            --app-primary-color:  var(--paper-indigo-500);

            /* Default paper custom properties */
            --primary-text-color: var(--paper-grey-900);
            --primary-background-color: #ffffff;
            --secondary-text-color: #737373;
            --disabled-text-color: #9b9b9b;
            --divider-color: #dbdbdb;
            --primary-color: var(--paper-indigo-500);
            --light-primary-color: var(--paper-indigo-100);
            --dark-primary-color: var(--paper-indigo-700);
            --accent-color: var(--paper-pink-a200);
            --light-accent-color: var(--paper-pink-a100);
            --dark-accent-color: var(--paper-pink-a400);
            }
        </style>
      </template>
    </dom-module>

I cannot make a PR because there is still #825 pending...

mercmobily:
@abdonrd  I agree. I updated the PR so that it does just that. It seems to make sense now...

robdodson:
We've totally revamped PSK2 so unfortunately this PR is no longer compatible.

I'm leaving open issues related to theming as we decide on what approach we want to use. Thanks for the PR @mercmobily. Coming up with theming guidance is the next p0 for us working on PSK

PR #825, opened 2016-05-08
 CSS reduction and add header effects

abdonrd:
Before:

may-08-2016 18-27-12

After:

may-08-2016 18-26-44

### DIFF
--- a/src/psk-app/psk-app.html
+++ b/src/psk-app/psk-app.html
@@ -49,6 +49,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         background-color: var(--paper-grey-50);
       }
 
+      a {
+        text-decoration: none;
+        color: inherit;
+        font-size: inherit;
+      }
+
       app-drawer {
         --app-drawer-content-container: {
           background-color: var(--app-primary-color);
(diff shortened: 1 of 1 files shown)


mercmobily:
Yes that's about 462804571 times better than my solution...

addyosmani:
LGTM

robdodson:
Thanks @abdonrd

abdonrd:
Yay! :)

robdodson:
I just noticed on desktop this looks pretty weird.

screen shot 2016-05-11 at 3 35 16 pm

joryphillips:
I gotta be honest with everyone. The shrink-on-scroll header title effect is the first thing I disable every time I launch a new project with PSK. _It is really cool_ — and to me seems more appropriate as a 'theme' than as the default configuration.

robdodson:
@joryphillips yeah on second thought I'm inclined to agree. We're going to have add back in CSS to fix this for desktop. May as well go with something simpler to begin with

robdodson:
Did some cleanup ➡️ https://github.com/PolymerElements/polymer-starter-kit/pull/828

abdonrd:
If we can get from [app-header/demo1.html](http://polymerlabs.github.io/app-layout/app-header/demo/demo1.html). Only need to add some more CSS to fix desktop view.

chuckh:
For mobile condensing the header makes better use of the screen real estate. I prefer fixing desktop over removing condensing.

abdonrd:
I'm agree with @chuckh! :)

robdodson:
I was thinking about this a bit in the car this morning. @joryphillips, @samccone, and myself seem to prefer not using the condensing effect and it sounds like @chuckh's concern is primarily around the use of space on mobile.

Would an acceptable compromise be to just not have a giant header, but instead always use a smaller, mobile friendly header. That way the user is likely not in the position of needing to remove a bunch of stuff if they don't like the header. Instead they can add effects if they want.

robdodson:
Basically this: https://polymerlabs.github.io/app-layout/templates/getting-started/

screen shot 2016-05-12 at 9 21 08 am

chuckh:
> Would an acceptable compromise be to just not have a giant header, but instead always use a smaller, mobile friendly header. 

@robdodson make sense to me!

abdonrd:
Makes sense to me too!

PR #828, opened 2016-05-12
Page cleanup

robdodson:
Reverting the condenses change in #825 to simplify things a bit.

Also got rid of the film and design page as they weren't really useful. We can illustrate the same thing with just two pages. Renamed the art page to about.


mercmobily:
How come you are no longer importing "../../bower_components/font-roboto/roboto.html" ?

robdodson:
@mercmobily I kind fo hating loading web fonts because they really screw up time to first paint.

mercmobily:
OK! I wondered because you aree still using Roboto in the style. So I wondered if it was going to work...

robdodson:
@mercmobily It'll eventually fall back to sans-serif if they don't have it installed already. We should probably throw Helvetica and Arial into the font stack

src/psk-app/psk-app.html
```diff
@@ -17,13 +17,11 @@
...
 
-<link rel="import" href="../../bower_components/font-roboto/roboto.html">
-
-<link rel="import" href="../../bower_components/paper-icon-button/paper-icon-button.html">
+<link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="../../bower_components/paper-styles/color.html">
 
 <link rel="import" href="../../bower_components/app-layout/app-drawer-layout/app-drawer-layout.html">
-<link rel="import" href="../../bower_components/app-layout/app-drawer/app-drawer.html">
+<link rel="import" href="../bower_components/app-layout/app-drawer/app-drawer.html">
```
abdonrd: Why this?

src/psk-app/psk-app.html
```diff
@@ -17,13 +17,11 @@
 <link rel="import" href="../../bower_components/iron-pages/iron-pages.html">
 <link rel="import" href="../../bower_components/iron-selector/iron-selector.html">
 
-<link rel="import" href="../../bower_components/font-roboto/roboto.html">
-
-<link rel="import" href="../../bower_components/paper-icon-button/paper-icon-button.html">
+<link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
```
abdonrd: Why this `../bower_components /` instead of `../../bower_components `?

index.html
```diff
@@ -28,7 +29,9 @@
 
     <link rel="import" href="/src/psk-app/psk-app.html">
   </head>
+
   <body>
+
```
abdonrd: Why these three new lines?

src/psk-app/psk-app.html
robdodson: ah just a mistake. I'll fix it.

index.html
robdodson: Just formatting the page a bit nicer

abdonrd: Personally, I prefer not have spaces or add another space before and after `</body>`.

PS: We need an style guide! 😃

addyosmani:
Overall this LGTM. I know there are one or two review nits above left to be addressed. 

On removing the Roboto import: if we were to experiment with the [Web Font Loader](https://github.com/typekit/webfontloader) library or Font Face Observer, it might be possible to come up with some smarter behaviour here that doesn't sacrifice much for first paint. I guess the only downside to not importing is a lot of our users (and their users) are going to see the fallback fonts rather than the version with Roboto. I'm fine with removal if you still get the Roboto import when trying to bring in the Paper elements.

robdodson:
OK fixed the messed up bower paths. Going to merge this so I can tackle the header compromise from https://github.com/PolymerElements/polymer-starter-kit/pull/825

PR #832, opened 2016-05-30
Simplify header

robdodson:
Creates a simplified mobile header as discussed here: https://github.com/PolymerElements/polymer-starter-kit/pull/825

screen shot 2016-05-30 at 1 33 03 pm


samccone:
+1

PR #844, opened 2016-06-05
Install only used components

chuckh:
Speeds up `polymer build` process from 58 seconds to 18 seconds. Will also speed deploying to http2 servers since unused components will not be copied to the server.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -3,8 +3,13 @@
   "dependencies": {
     "app-layout": "PolymerElements/app-layout#^0.9.0",
     "app-route": "PolymerElements/app-route#^0.9.1",
-    "iron-elements": "PolymerElements/iron-elements#^1.0.0",
-    "paper-elements": "PolymerElements/paper-elements#^1.0.0",
+    "iron-flex-layout": "PolymerElements/iron-flex-layout#^1.0.0",
+    "iron-icon": "PolymerElements/iron-icon#^1.0.0",
+    "iron-iconset-svg": "PolymerElements/iron-iconset-svg#^1.0.0",
+    "iron-pages": "PolymerElements/iron-pages#^1.0.0",
+    "iron-selector": "PolymerElements/iron-selector#^1.0.0",
+    "paper-icon-button": "PolymerElements/paper-icon-button#^1.0.0",
+    "paper-styles": "PolymerElements/paper-styles#^1.0.0",
     "polymer": "Polymer/polymer#^1.0.0"
   },
   "private": true


abdonrd:
LGTM!

TimvdLippe:
While fewer components is nice, I think that leaving the elements collections in is friendlier for new developers trying out Polymer. The transition to and tinkering with different `iron-*` or `paper-*` elements is a lot easier and smoother.

Shouldn't `polymer build` be able to figure out which files it requires by analyzing HTML imports (and possibly some lazy-loaded onces)?

addyosmani:
Thanks for the PR, @chuckh! My two cents are that @TimvdLippe has a good point here. When you're prototyping in particular, it's nice having access to the complete set of elements for those collections. I could imagine the time-savings for the average project being negligible.

Does the impact of installing the extra pieces slow us down pretty considerably?

chuckh:
@addyosmani I did this pull request because of the issues opened in PSK1 regarding initial install time and deploy time.  The initial time to do a `bower install` is less then half the time, but the current way is not that bad. This more effects the ongoing `polymer build` time and deploying to servers such as Firebase Hosting. When deploying PSK2 to firebase it went from 3 minutes and 58 seconds to 57 seconds. Firebase deploy went from copying 2700 files to 126 files.

I agree @TimvdLippe point also. We could use this method and add instructions about how to add addition components in the readme or set expectations in readme for install, build and deploy times.

I open an issue on Polymer-CLI hopefully to better handle the build process, see https://github.com/PolymerElements/polymer-starter-kit/pull/845.  If this gets address it makes this change unnecessary.

abdonrd:
I think people start to download and import the components is also important.

joryphillips:
When I went through the Polymer onboarding process, all the documentation and instructions I saw everywhere emphasized the need to download the individual components that I needed. Eventually I realized I was needlessly, repeatedly downloading elements. 

Here's a vote for discouraging bloat AND repetitive needlessness with clearer documentation of what gets installed by default and what doesn't. 

@robdodson will there continue to be a 'beginner' and 'intermediate' version of PSK?

robdodson:
> @robdodson will there continue to be a 'beginner' and 'intermediate' version of PSK?

I don't think so as the main difference between the two is the gulp file and that's going away in PSK2 to be replaced by the CLI

abdonrd:
New about this? I'm in favor of merge the PR.

robdodson:
Thanks @chuckh. Sounds like cutting our install time will improve the experience for folks so +1 :)

PR #846, opened 2016-06-06
add updated tutorials from polymer-project.org

kaycebasques:
The PSK docs are no longer hosted on polymer-project.org, so I'm moving them over here so that they can live on. I also updated the deploy doc to fix the issues mentioned in https://github.com/Polymer/docs/issues/1629 

cc @arthurevans @dandv


docs/tutorials/deploy.md
```diff
@@ -0,0 +1,54 @@
...
+    select `No`.
+
+1.  Select which app you want to deploy to. 
+
+1.  Run the following command to deploy.
+
+        firebase deploy
+
+1.  You can view your site at the URL listed in the output, or by running 
+    `firebase open` and then selecting `Hosting: Deployed Site`.
```
chuckh: You can use `firebase open hosting:site` which will open the site without having to make a selection.

chuckh:
Overall LGTM, just one minor suggestion.

addyosmani:
LGTM. Can we make @chuckh's suggestion above?

kaycebasques:
Done! Thanks for reviews @chuckh and @addyosmani

PR #848, opened 2016-06-07
Fixes #847, fixes #817

### LINKED ISSUE #817: psk should be dropped from psk-element-name for psk2
I was considering starting an app with psk2, and I noticed I'll have to change all the psk-element names. For example, I'll have to change `psk-page-404` to `page-404`. If this project is suppose to be a starting point or template for projects, I think it'd be good if psk was dropped from the element names by default. It would make it easier to start new projects from this project.

Thoughts?

### LINKED ISSUE #847: PSK2: Reorganizing Elements by Functionality and Relationship
### Situation
Right now every element, regardless of function, is placed inside the src directory.

    src/psk-app/psk-app.html
    src/psk-page-home/psk-page-home.html
    src/psk-page-about/psk-page-about.html
    src/psk-page-404/psk-page-404.html

### Problem
The problem with this approach is, that though the elements belong together in a hierarchical way (f.e. psk-page-home is inside psk-app), this hierarchical structure is not reflected in the structure of the app, thus the connection between elements has to be understood by the "name" of each element.

This leads to the elements needing a descriptive and thus longer name, which leads to "three-word-elements". In our case "psk-page-about" or "psk-page-404".

### Implication
The implicated results are, that as the app grows, the list of elements that are placed inside the src becomes longer. Thus making it more difficult to understand the relationship between the elements, by looking into the directory. This might even make "four-word-element-names" necessary. Who knows...

### Suggestion
By structuring the psk-app, by adding a views directory to it and moving the "pages" inside this views directory, like this:

    src/psk-app/psk-app.html
    src/psk-app/views/psk-home/psk-home.html
    src/psk-app/views/psk-about/psk-about.html

#### Benefits of this approach is: 
- Function and relationship between the elements become more intuitive, because of separation
- The two-word naming convention for custom elements can be kept, because of directory name, thus also solving Issue [#817](https://github.com/PolymerElements/polymer-starter-kit/issues/817)
- Increased clarity, maintainability and intuitiveness.

LoveAndHappiness:
Fixes #847 by moving elements from src to psk-app/views to make the elements hierarchy be reflected by the directory structure. 

Fixes #817 by allowing for the custom elements to follow the "two-word" naming convention.

### DIFF
--- a/src/psk-app/psk-app.html
+++ b/src/psk-app/psk-app.html
@@ -29,9 +29,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../../bower_components/app-layout/app-toolbar/app-toolbar.html">
 
 <link rel="import" href="../app-icons/app-icons.html">
-<link rel="import" href="../psk-page-home/psk-page-home.html">
-<link rel="import" href="../psk-page-about/psk-page-about.html">
-<link rel="import" href="../psk-page-404/psk-page-404.html">
+<link rel="import" href="views/psk-home/psk-home.html">
+<link rel="import" href="views/psk-about/psk-about.html">
+<link rel="import" href="views/psk-404/psk-404.html">
 
 <dom-module id="psk-app">
   <template>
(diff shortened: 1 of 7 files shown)


LoveAndHappiness:
I signed it!

stramel:
@robdodson what is the status on this?

robdodson:
I don't think we're going to go with this approach. Discussed in https://github.com/PolymerElements/polymer-starter-kit/issues/847

PR #854, opened 2016-06-22
Upd dep's, Polymer 1.5, WCT 4.3.0, PE 2.0.0

arturparkhisenko:


abdonrd:
Closed in favor of [PSK2](https://github.com/PolymerElements/polymer-starter-kit/blob/psk2/bower.json).

PR #857, opened 2016-06-29
Psk2 polymer build

robdodson:
@abdonrd @chuckh @addyosmani @justinfagnani PTAL

Currently when I run this it seems to work but the gulp process logs:

```
Ignoring duplicate element definition: undefined
Ignoring duplicate element definition: undefined
Ignoring duplicate element definition: undefined
Ignoring duplicate element definition: undefined
Ignoring duplicate element definition: undefined
Ignoring duplicate element definition: undefined
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -24,11 +24,8 @@ const polymer = require('polymer-build');
 const PolymerProject = polymer.PolymerProject;
 const fork = polymer.forkStream;
 
-let project = new PolymerProject({
-  root: process.cwd(),
-  entrypoint: 'index.html',
-  shell: 'src/psk-app.html',
-});
+let polymerJSON = require('./polymer.json');
+let project = new PolymerProject(polymerJSON);
 
 gulp.task('default', () => {
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,3 +1,13 @@
+/*
+@license
+Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+*/
+
 'use strict';
 
 const gulp = require('gulp');
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,6 +1,5 @@
 const gulp = require('gulp');
 const gulpif = require('gulp-if');
-const babel = require('gulp-babel');
 const imagemin = require('gulp-imagemin');
 const logging = require('plylog');
 const mergeStream = require('merge-stream');
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,3 +1,5 @@
+'use strict';
+
 const gulp = require('gulp');
 const gulpif = require('gulp-if');
 const imagemin = require('gulp-imagemin');
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,0 +1,58 @@
+const gulp = require('gulp');
+const gulpif = require('gulp-if');
+const babel = require('gulp-babel');
+const imagemin = require('gulp-imagemin');
+const logging = require('plylog');
+const mergeStream = require('merge-stream');
+
+const polymer = require('polymer-build');
+
+// Got problems? Try logging 'em
+// logging.setVerbose();
+
+const PolymerProject = polymer.PolymerProject;
+const fork = polymer.forkStream;
+
+let project = new PolymerProject({
+  root: process.cwd(),
+  entrypoint: 'index.html',
+  shell: 'src/psk-app.html',
+});
+
+gulp.task('default', () => {
+
+  // process source files in the project
+  let sources = project.sources()
+    .pipe(project.splitHtml())
+    // add compilers or optimizers here!
+    .pipe(gulpif('*.js', babel({
+      presets: ['es2015']
+    })))
+    .pipe(gulpif('*.{png,gif,jpg,svg}', imagemin({
+      progressive: true, interlaced: true
+    })))
+    .pipe(project.rejoinHtml());
+
+  // process dependencies
+  let dependencies = project.dependencies()
+    .pipe(project.splitHtml())
+    // add compilers or optimizers here!
+    .pipe(project.rejoinHtml());
+
+  // merge the source and dependencies streams to we can analyze the project
+  let allFiles = mergeStream(sources, dependencies)
+    .pipe(project.analyze);
+
+  // fork the stream in case downstream transformers mutate the files
+  // this fork will vulcanize the project
+  let bundled = fork(allFiles)
+    .pipe(project.bundle)
+    // write to the bundled folder
+    .pipe(gulp.dest('build/bundled'));
+
+  let unbundled = fork(allFiles)
+    // write to the unbundled folder
+    .pipe(gulp.dest('build/unbundled'));
+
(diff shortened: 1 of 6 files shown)


gulpfile.js
```diff
@@ -0,0 +1,58 @@
...
+
+const polymer = require('polymer-build');
+
+// Got problems? Try logging 'em
+// logging.setVerbose();
+
+const PolymerProject = polymer.PolymerProject;
+const fork = polymer.forkStream;
+
+let project = new PolymerProject({
```
moderndeveloperllc: Could this use the `polymer.json` file?

robdodson: good question, @justinfagnani wdyt?

jsilvermist:
Aren't the 6 errors because the elements, `psk-page-404.html`, `psk-page-about.html`, `psk-page-home.html`, all don't have any `is` defined in the call to polymer, and they are being processed twice for bundled and unbundled?

```
class PSKPageAbout {

}

Polymer(PSKPageAbout);
```

If I'm completely wrong I apologize, but should this not be:

```
class PSKPageAbout {

  beforeRegister() {
    this.is = 'psk-page-about';
  }

}

Polymer(PSKPageAbout);
```

robdodson:
@jsilvermist ah good catch! Let's see if that fixes things. cc @justinfagnani as an fyi

robdodson:
Hm seems like i'm still getting those duplicate definition errors even after adding `is=`

TimvdLippe:
@robdodson Maybe the missing `extends HTMLElement` on the ES6 classes?

robdodson:
@TimvdLippe oooooh maybe. let me try that

robdodson:
@TimvdLippe still seeing the error but good catch on the missing `extends` I've updated the PR to include it.

TimvdLippe:
I can take a look tomorrow and debug the error. The missing `extends` is probably a good candidate for the Polylinter or a warning in the core library (although previous discussions indicate it has a low chance in merging).

jsilvermist:
What is the benefit to using classing in this way? It seems to just complicate things, and without even reusing classes across elements, what's the real benefit verses just using an object like before?

robdodson:
@jsilvermist It's more inline with how you create a vanilla Custom Element and Polymer team is thinking of adding class syntax to a future release.

Because CLI can use multiple templates I think that the current [app-drawer-template](https://github.com/Polymer/polymer-cli/issues/226#issuecomment-224571538) will become the new "basic" template and folks who don't want to use classes or ES6 can use that.

PSK will become an example that demonstrates how to add in features that CLI doesn't offer (image optimization, Babel, etc). A lot of developers have requested ES2015 support so this template becomes the logical place to do that.

Then there will be other templates which demonstrate other concepts (auth with firebase, typescript, etc).

jsilvermist:
@robdodson Ah I see, that makes sense. I was wondering how the psk would fit in with all the templates in the CLI.

abdonrd:
It works for me!

src/psk-app.html
```diff
@@ -146,22 +146,33 @@ <h1 title>Polymer Starter Kit</h1>
...
+          foo: {
+            type: String,
+            value: 'Hello, World'
+          }
+        };
+      }
 
-      attached: function() {
+      attached() {
         if (!this.route.path) {
```
abdonrd: I think we can remove this. @rictic can you confirm, please? :)

gulpfile.js
abdonrd: Makes sense!

abdonrd:
@robdodson why not have a `gulp-tasks` folder with the tasks? :)

robdodson:
It sounds like the team is planning to tweak Polymer core to make it easier to register ES6 classes. Rather than cause a bunch of confusion transitioning from one ES6 syntax to another (and waiting on the new syntax to land) I think I'd like to swap this back to the current ES5 style and just ship the thing

chuckh:
> I think I'd like to swap this back to the current ES5 style and just ship the thing

@robdodson I agree. Rather do this once too!

abdonrd:
Makes sense!

gulpfile.js
```diff
@@ -0,0 +1,55 @@
+const gulp = require('gulp');
+const gulpif = require('gulp-if');
+const babel = require('gulp-babel');
```
abdonrd: Remove :)

package.json
```diff
@@ -12,5 +12,14 @@
   "engines": {
     "node": ">=0.10.0"
   },
-  "private": true
+  "private": true,
+  "devDependencies": {
+    "babel-preset-es2015": "^6.9.0",
+    "gulp-babel": "^6.1.2",
```
abdonrd: Remove :)

robdodson:
OK I updated the PR so it's back to ES5. Anyone care to take a look?

package.json
```diff
@@ -12,5 +12,14 @@
   "engines": {
     "node": ">=0.10.0"
   },
-  "private": true
+  "private": true,
+  "devDependencies": {
+    "babel-preset-es2015": "^6.9.0",
```
abdonrd: Remove :)

robdodson: good call. Removed that and other references in the gulpfile. PTAL

abdonrd: I can still see the babel dependencies. 😕

robdodson: doh! how about now :P

abdonrd: Yay! 🎉

gulpfile.js
```diff
@@ -0,0 +1,54 @@
+const gulp = require('gulp');
+const gulpif = require('gulp-if');
+const imagemin = require('gulp-imagemin');
+const logging = require('plylog');
```
abdonrd: @robdodson We aren't using this.

robdodson: It's commented out but it's there in case the user needs it
https://github.com/PolymerElements/polymer-starter-kit/pull/857/files#diff-b9e12334e9eafd8341a6107dd98510c9R9

abdonrd:
We can use `const` in the `gulpfile.js` with node `>=4.0`?

robdodson:
@abdonrd goooood question...

gulpfile.js
abdonrd: Oops, right!

robdodson:
@abdonrd it looks like we can use const/let in gulp 4 if I add `use strict` which I did. Also added a local dependency for gulp since that was missing :P

abdonrd:
@robdodson nice! But... Gulp 4? I was talking about Node 4. :)

gulpfile.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
```
abdonrd: We have forgotten the license.

robdodson:
Bad commit message I meant Node 4. I tested with Node 4.4.7

robdodson:
ok added license

abdonrd:
Excluding [#857/files#r69175814](https://github.com/PolymerElements/polymer-starter-kit/pull/857/files#r69175814), LGTM! 🎉

robdodson:
@abdonrd added a line to the gulpfile to pull in polymer.json that should address #857/files#r69175814. Spoke with the CLI team and this seems to be the recommendation. We should be ok to drop the `root` property because gulp sets it to the right location for us

robdodson:
I'm gonna go ahead and merge so we can keep moving forward 🚀

jsilvermist:
Why remove Babel just because we have to wait longer for official class syntax?

Wouldn't it make more sense to keep Babel in by default for use with other JS, and then just convert Polymer calls from objects to classes later on?

abdonrd:
@robdodson nice work Rob! 💪

robdodson:
@jsilvermist I think we'll switch back to using it once we have proper class syntax, in the meantime it's easy enough for people to add it themselves. I want to avoid where we ended up with the previous PSK where we had every gulp plugin in the world installed and spent most of our time just helping people debug the gulpfile.

PR #865, opened 2016-07-18
Looks better to have a close icon to close toast

jtvjan:
This may look better than it does now, I already tested it so it works.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -180,7 +180,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     </paper-drawer-panel>
 
     <paper-toast id="toast">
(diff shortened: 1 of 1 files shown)


jtvjan:
I signed it!

robdodson:
Hey @jtvjan thanks for the pull request! We're using the Material Design guidelines to dictate our usage of components and they mention that [snackbars/toasts should not contain icons](https://material.google.com/components/snackbars-toasts.html). I think we're going to stick with that guidance for now but thank you again for taking the time to submit the change :)

jtvjan:
Ok. But the '0-1 actions, not dismiss or cancel' section in that very document contradicts the OK button. Not saying that my PR is better, just saying that you might want to remove the OK button altogether.

robdodson:
Hm good point.

jtvjan:
I made that change I said in my previous comment, see #866

PR #866, opened 2016-07-18
Change made to confirm to MD specs

jtvjan:
This part of the MD specifications says that you shouldn't have dismiss buttons: https://material.google.com/components/snackbars-toasts.html#snackbars-toasts-usage
Continuing PR #865
I tested it and it works. As far as I know there are no toasts without timeout in PSK.


robdodson:
Ah looks like we need to fix our failing Firefox tests. No big, this LGTM

PR #871, opened 2016-07-20
Fix app-layout breaking change: title > main-title

abdonrd:
Reference: https://github.com/PolymerElements/app-layout/releases/tag/v0.10.0


robdodson:
Nice catch! thanks!

PR #877, opened 2016-07-26
role="main" to iron-pages

abdonrd:


src/psk-app.html
```diff
@@ -115,8 +115,11 @@
           </app-header>
 
           <!-- drawer menu -->
-          <iron-selector selected="[[page]]" attr-for-selected="name"
-            on-iron-activate="_drawerSelected" role="navigation">
+          <iron-selector
+              selected="[[selectedPage]]"
```
chuckh: We should try to stay consist with app-drawer-template if possible.  See https://github.com/Polymer/app-drawer-template/blob/master/src/my-app.html#L77. They use page instead of selectedPage.

abdonrd: I'm agree with consist!
But we can add a PR there? The same with the 404 we have seen today. WDYT @chuckh?

chuckh: I like `selectedPage` over `page` but it may not be worth the effort to change in all the places is is used/shown (app-drawer-template, the doc at https://www.polymer-project.org/1.0/start/toolbox/create-a-page, etc.).

abdonrd: Okey! Go back!

PR #887, opened 2016-07-28
Add aria live region to inform when page has changed

felixzapata:
Based on the idea of the Polymer Shop App, I´ve added an aria live region to inform when the page has changed. 

This announcer can be used any time that the app needs to inform about something that happened.

### DIFF
--- a/src/psk-app.html
+++ b/src/psk-app.html
@@ -191,8 +191,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       _pageChanged: function(page) {
         // Load page import on demand. Show 404 page if fails
         var resolvedPageUrl = this.resolveUrl('psk-page-' + page + '.html');
-        this._announce('Page ' + page + ' loaded');
-        this.importHref(resolvedPageUrl, null, this._showPage404, true);
+        this.importHref(resolvedPageUrl, this._announce('Page ' + page + ' loaded'), this._showPage404, true);
       },
 
       _showPage404: function() {
--- a/src/psk-app.html
+++ b/src/psk-app.html
@@ -191,7 +191,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       _pageChanged: function(page) {
         // Load page import on demand. Show 404 page if fails
         var resolvedPageUrl = this.resolveUrl('psk-page-' + page + '.html');
-        this.importHref(resolvedPageUrl, this._announce('Page ' + page + ' loaded'), this._showPage404, true);
+        var onload = function() {
+          this._announce('Page ' + page + ' loaded');
+        };
+        this.importHref(resolvedPageUrl, onload, this._showPage404, true);
       },
 
       _showPage404: function() {
--- a/src/psk-app.html
+++ b/src/psk-app.html
@@ -144,6 +144,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       <!-- main menu -->
       <iron-pages
+          selected-attribute="visible"
           selected="[[page]]"
           attr-for-selected="name"
           fallback-selection="404"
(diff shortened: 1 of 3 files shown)


src/psk-app.html
```diff
@@ -164,13 +176,22 @@ <h1 main-title>Polymer Starter Kit</h1>
...
+
       _routePageChanged: function(page) {
         this.page = page || 'home';
       },
 
       _pageChanged: function(page) {
         // Load page import on demand. Show 404 page if fails
-        var resolvedPageUrl = this.resolveUrl('psk-page-' + page + '.html')
+        var resolvedPageUrl = this.resolveUrl('psk-page-' + page + '.html');
+        this._announce('Page ' + page + ' loaded');
```
robdodson: this won't have loaded until after line 195 has run (or it might not load at all if there is an error).

the 2nd argument to importHref takes an onload function (https://www.polymer-project.org/1.0/docs/api/Polymer.Base#method-importHref) so you'd want to call _announce from in there

felixzapata: @robdodson done

src/psk-app.html
```diff
@@ -191,8 +191,7 @@ <h1 main-title>Polymer Starter Kit</h1>
       _pageChanged: function(page) {
         // Load page import on demand. Show 404 page if fails
         var resolvedPageUrl = this.resolveUrl('psk-page-' + page + '.html');
-        this._announce('Page ' + page + ' loaded');
-        this.importHref(resolvedPageUrl, null, this._showPage404, true);
+        this.importHref(resolvedPageUrl, this._announce('Page ' + page + ' loaded'), this._showPage404, true);
```
robdodson: i don't think this will work. I think you will need to put an actual function declaration here. Something like

this.importHref(resolvedPageUrl, function() {
  this._announce('Page ' + page + ' loaded')
}, this._showPage404, true);

felixzapata: ok, I will check it again. Sorry

felixzapata: done

robdodson:
@abdonrd @chuckh either of you have opinions on this approach?

chuckh:
@robdodson looks promising. I need to test it some before giving strong opinion. I won't have a chance until next week to do so.

src/psk-app.html
```diff
@@ -93,6 +93,13 @@
       iron-pages > * {
         min-height: 100vh;
       }
+
+      .announcer {
+        position: fixed;
```
abdonrd: Why the position fixed it is hidden?

src/psk-app.html
```diff
@@ -164,14 +176,25 @@ <h1 main-title>Polymer Starter Kit</h1>
         '_routePageChanged(routeData.page)'
       ],
 
+      // A11y announce the given message.
+      _announce: function(message) {
+        this._a11yLabel = '';
```
abdonrd: Why clean the message?

abdonrd:
Maybe we should use the [`iron-a11y-announcer`](https://elements.polymer-project.org/elements/iron-a11y-announcer)?

src/psk-app.html
felixzapata: I think is not necessary. At this moment I´ve used the same styles that in Polymer Shop.

src/psk-app.html
felixzapata: To reset the message and to be sure that is has not previous messages. Although you set the message and this case it would be very strange. So, maybe it is not necessary

felixzapata:
Yes, it can be use this component. For the moment, I used the same as the Polymer Shop but I can change to use the componet.

abdonrd:
I think we can update the PR with the [`iron-a11y-announcer`](https://elements.polymer-project.org/elements/iron-a11y-announcer). 😁

Then we can remove the CSS and the HTML.

abdonrd:
And what happens if the name of the page has more than one word?
"Page page-with-more-than-one-word loaded" 😕

felixzapata:
@abdonrd for this kind of page name, the name should be "sanitized" before using it in the announcer.
 
But, after thinking about that, it should not be the page name because maybe the name is totally a nonsense of words to inform the user. There should be a map or something with the relation between the page names and the real name of the pages to inform. Or, when the page has changed, try to look for the real name of the page using the route.

felixzapata:
@abdonrd I've updated the pull request using the same technique that in the Polymer Shop. 

Every page fires an event when is visible. The app component has a listener to this event and call the announcer with the correct title.

Anyway, it can obtain the same result using another system. Maybe you can listen another event. For example, the `iron-select` event. The problem with this event is you can listen other components you don't want. 

Or maybe you can pass to `_pageChanged` the real name of the page.

abdonrd:
Perhaps this adds too much complexity. 😕
WDYT @robdodson?

robdodson:
I think we may want to move it over to the custom build seed we're working on.

felixzapata:
Have i to do something?

robdodson:
@felixzapata not right now. We're still sorting things out

robdodson:
I think we may want to see how we can implement this perhaps without using an aria live region. It may be a better experience to move focus to the heading for the section that just loaded so the user has an efficient keyboarding experience.

felixzapata:
ok, I will create a new pull request.

PR #893, opened 2016-08-05
Add 404 to fragments and tweak readme to help point this out

robdodson:


robdodson:
@chuckh PTAL

polymer.json
```diff
@@ -4,6 +4,7 @@
   "fragments": [
     "src/my-view1.html",
     "src/my-view2.html",
-    "src/my-view3.html"
+    "src/my-view3.html",
+    "src/my-view404.html"
   ]
 }
```
chuckh: Do we also need to add:

"includeDependencies": [
    "bower_components/webcomponentsjs/webcomponents-lite.min.js"
  ]

see https://github.com/PolymerElements/polymer-starter-kit/blob/psk2/polymer.json

chuckh: We added it to PSK2 because it is lazy loaded.  See https://github.com/PolymerElements/polymer-starter-kit/pull/889.

robdodson:
@FredKSchott I've added `includeDependencies` for webcomponents-lite.js but it doesn't seem to be picked up. Doing `polymer build` and `polymer serve build/bundled` gives a blank page in Firefox. I can see the request going out for webcomponents-lite.js but it isn't copied to the build dir so the server just sends back index.html

abdonrd:
@robdodson: @FredKSchott said me the `polymer-cli` don't get the `includeDependencies`. But the `polymer-build` get it.

chuckh:
LGTM!

PR #895, opened 2016-08-09
Simplify wct.conf.js

ankon:
The magic for bower_components isn't needed, as bower_components is already under 'app' now.

----
Note that even if that path mapping would be needed, it would be simpler to define it directly:
~~~~
 var ret = {
   'suites': ['app/test'],
   'webserver': {
     'pathMappings': [
       { '/components/<basename>/app/bower_components': 'bower_components' }
     ]
   }
 };
~~~~

With this change one could even consider moving from a `wct.conf.js` to a `wct.conf.json`, but that makes adding comments harder -- which is likely what happens when this is used to create a new more complex app.


robdodson:
We've removed wct.conf.js from PSK as we're trying to keep the project as low overhead as possible.

PR #897, opened 2016-08-10
Update build commands

abdonrd:
Short and avoid comments in the 'terminal' code as the 'Start the development server' step.


robdodson:
Instead of PR'ing against `release`, would you mind doing it against the app-drawer-migration branch?

PR #908, opened 2016-08-29
Set focus on title when page changes

felixzapata:
The pages use the  `selected-attribute` of `iron-pages` to set the focus on the title when the pages have changed.

This pull request is related with [this one](https://github.com/PolymerElements/polymer-starter-kit/pull/887).

### DIFF
--- a/src/manage-focus-behavior.html
+++ b/src/manage-focus-behavior.html
@@ -13,10 +13,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <script>
   Polymer.ManageFocusBehavior = {
     properties: {
-        visible: {
-          type: Boolean,
-          observer: '_visibleChanged'
-        }
+      visible: {
+        type: Boolean,
+        observer: '_visibleChanged'
+      }
     },
     _visibleChanged: function(visible) {
       var title = Polymer.dom(this.root).querySelector('h1'); 
--- a/src/manage-focus-behavior.html
+++ b/src/manage-focus-behavior.html
@@ -11,20 +11,18 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../bower_components/polymer/polymer.html">
 
 <script>
-  (function() {
-    'use strict';
-    Polymer.MyViewBehavior = {
-      properties: {
-         visible: {
-           type: Boolean,
-           observer: '_visibleChanged'
-         }
-      },
-      _visibleChanged: function(visible) {
-        if (visible) {
-          Polymer.dom(this.root).querySelector('h1').focus();
+  Polymer.ManageFocusBehavior = {
+    properties: {
+        visible: {
+          type: Boolean,
+          observer: '_visibleChanged'
         }
+    },
+    _visibleChanged: function(visible) {
+      var title = Polymer.dom(this.root).querySelector('h1'); 
+      if (visible && title) {
+        title.focus();
       }
-    };
-  })();
+    }
+  };
 </script>
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -93,6 +93,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
         <iron-pages
             selected="[[page]]"
+            selected-attribute="visible",
             attr-for-selected="name"
             fallback-selection="view404"
             role="main">
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -21,7 +21,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../bower_components/iron-selector/iron-selector.html">
 <link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="my-icons.html">
-<link rel="import" href="my-view-behavior.html">
+<link rel="import" href="manage-focus-behavior.html">
 
 <dom-module id="my-app">
   <template>
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -21,6 +21,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../bower_components/iron-selector/iron-selector.html">
 <link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="my-icons.html">
+<link rel="import" href="my-view-behavior.html">
 
 <dom-module id="my-app">
   <template>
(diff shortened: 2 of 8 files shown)


src/my-view2.html
```diff
@@ -23,15 +23,26 @@
...
       <p>Ea duis bonorum nec, falli paulo aliquid ei eum.</p>
       <p>Id nam odio natum malorum, tibique copiosae expetenda mel ea.Detracto suavitate repudiandae no eum. Id adhuc minim soluta nam.Id nam odio natum malorum, tibique copiosae expetenda mel ea.</p>
     </div>
   </template>
 
   <script>
     Polymer({
-      is: 'my-view2'
+      is: 'my-view2',
+      properties: {
```
TimvdLippe: Instead of copy-pasting the functions and properties, I think it is better to put them in a behavior `MyViewBehavior`.

felixzapata: ok, I will change it.

src/my-app.html
```diff
@@ -21,6 +21,7 @@
 <link rel="import" href="../bower_components/iron-selector/iron-selector.html">
 <link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="my-icons.html">
+<link rel="import" href="my-view-behavior.html">
```
robdodson: suggest renaming this something like `manage-focus-behavior`

src/my-view-behavior.html
```diff
@@ -0,0 +1,30 @@
...
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<link rel="import" href="../bower_components/polymer/polymer.html">
+
+<script>
+  (function() {
```
robdodson: can remove the self executing function and usage of `use strict`. It doesn't match the other elements and I don't think we're planning to swap over to this style

src/my-view-behavior.html
```diff
@@ -0,0 +1,30 @@
...
+    'use strict';
+    Polymer.MyViewBehavior = {
```
robdodson: Polymer.ManageFocusBehavior

src/my-view-behavior.html
```diff
@@ -0,0 +1,30 @@
...
+      properties: {
+         visible: {
+           type: Boolean,
+           observer: '_visibleChanged'
+         }
+      },
+      _visibleChanged: function(visible) {
+        if (visible) {
+          Polymer.dom(this.root).querySelector('h1').focus();
```
robdodson: Should check that h1 actually exists before calling focus on it, otherwise if they remove the h1 it'll throw an error

robdodson: I think we'll also need a style in the shared-styles to set `h1` outline to `none`

felixzapata:
hi @robdodson , I´ve updated the pull request.

src/manage-focus-behavior.html
```diff
@@ -0,0 +1,28 @@
...
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<link rel="import" href="../bower_components/polymer/polymer.html">
+
+<script>
+  Polymer.ManageFocusBehavior = {
+    properties: {
+        visible: {
```
robdodson: indentation seems off. should be 2 spaces

felixzapata: done

robdodson:
Sorry for the delay on this...

After thinking about this PR for a bit, I still don't feel like we've hit upon the best solution. My concern is that adding a behavior, which queries for a specific element, creates an implicit dependency which most developers will miss. If they remove the h1 from many of their pages then they're loading a behavior that doesn't really do anything, it's just dead code at that point.

I was wondering if it might be better to instead blur() whenever they click a menu item. This would make the behavior of the page similar to that of a classic, multipage website, where each time you click a link, focus is reset to the top of the document. This would reduce the code down to a one liner and there would no longer be any implicit dependencies. The problem with this approach is it doesn't seem to work, at least not in Chrome :P  Calling `document.activeElement.blur()` does reset focus to `<body>` but the focus navigation start point seems to remain where it was. So if the user was previously focused on the `View 1` anchor, and you blur, the next time they press tab focus will jump to the `View 2` anchor. Also I found that in VoiceOver, calling `blur()` did not start reading from the top of the page as I had hoped. But I'd like to keep exploring this approach.

Since this is a generic starting point, I'd like any focus management to be very obvious and not create any dependencies which the developer might accidentally break in the natural course of adding their own content. I'll try to bug some of the accessibility folks on Chrome to see if they have any ideas.

felixzapata:
Maybe it is related about the bug you comment in Chrome but,  a few weeks ago I´ve discovered that if you set a focus (or a blur) event to the document and the document uses components with Shadow DOM, the listener does not work correctly. But, if you uses the Shady DOM or the Polyfill, the listener works.

A similar problem happens with the ChromeVox extension. The focus when you press the key tab does not work if you uses Shadow DOM. 

I think the issue is because of how Shadow DOM v0 manages these events. I read about (these events with Shadow DOM v1](https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom#focus) and they work better.

felixzapata:
hi @robdodson, any news about that? Maybe at the moment we can add a basic message to the user when the route changes. 

The original problem was that there was no additional inform to compose a message like "View One loaded". Instead, the basic could be "The page has changed".

TimvdLippe:
This PR has been open for quite a while, but I am echoing the comment of @robdodson at https://github.com/PolymerElements/polymer-starter-kit/pull/908#issuecomment-254041025 While I do think that title management is important, this implementation seems to create a hidden dependency that is really easy to trip over by new users.

I am not sure what other options we have to not introduce magic, but keep it DRY. @felixzapata maybe you have an idea to achieve this, else I think we are inclined to close this PR per the current implementation.

robdodson:
I think this is one of those places where we are just lacking good accessibility APIs. I would personally like an API to move the [focus navigation start point](https://developers.google.com/web/updates/2016/03/focus-start-point). This is something @alice and I have discussed a few times, but haven't had the chance to try to spec it out.

TimvdLippe:
Closing this PR for now, while focus navigation management still needs to resolved. Thanks @felixzapata for your effort and sorry for not being able to incorporate these changes at this moment in time.

PR #910, opened 2016-08-29
✂️

notwaldorf:
PTAL 👀  @robdodson @azakus


azakus:
ISSUE TEMPLATE should stick around. LGTM otherwise.

notwaldorf:
@azakus Done!

abdonrd:
Why remove the licenses?

notwaldorf:
@abdonrd This is what `polymer init starter-kit` generates, which is the user's brand new project, that we didn't write, so _our_ license shouldn't be on it by default

notwaldorf:
Ok after more talking, we should leave the source licenses in and just remove the unused license.txt, so the user can provide their own for new changes. Licenses are hard.

robdodson:
Thanks @notwaldorf !

jsilvermist:
Aww sad =/ I know the license in the files have prevented some groups and companies from being allowed to use it in the past.

PR #928, opened 2016-10-05
Set firefox to latest for travis

chuckh:
as Selenium - Firefox issue is now fixed SeleniumHQ/selenium#2110.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 sudo: required
 dist: trusty
 addons:
-  firefox: '47.0.1'
+  firefox: latest
   apt:
     sources:
       - google-chrome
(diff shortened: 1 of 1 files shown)


robdodson:
Cool. Feels like we should add firefox to the test matrix. I think right now it's only running tests against chrome

chuckh:
So for now we need to leave `firefox: `46` otherwise tests will not run. I think it is related to this issue: https://github.com/Polymer/web-component-tester/issues/367.

robdodson:
Looks like Polymer/web-component-tester#367. Maybe we can try again with the latest FF?

chuckh:
@robdodson changed firefox to latest and tests passed!

robdodson:
awwwwwww yissssssss

PR #932, opened 2016-10-13
Drawer will hide on page change, unless drawer is persistent

mercmobily:
Hi,

The title says it all!
It's only has minor modifications, but with great gains.
I gave an ID to the drawer so that I could refer to it with `this.$.drawer`. I hope that's accetable!

Merc.

### DIFF
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -123,8 +123,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _routePageChanged: function(page) {
         this.page = page || 'view1';
-
-        if (!this.$.drawer.persistent) this.$.drawer.opened = false;
+        if (!this.$.drawer.persistent) this.$.drawer.close();
       },
 
       _pageChanged: function(page) {
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -72,7 +72,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     <app-drawer-layout fullbleed>
       <!-- Drawer content -->
-      <app-drawer id="drawer" opened="{{drawerOpened}}">
+      <app-drawer id="drawer">
         <app-toolbar>Menu</app-toolbar>
         <iron-selector selected="[[page]]" attr-for-selected="name" class="drawer-list" role="navigation">
           <a name="view1" href="/view1">View One</a>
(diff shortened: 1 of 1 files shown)


jsilvermist:
Why use a (2 way) binding, why not just set `this.$.drawer.opened = false;`?

mercmobily:
@jsilvermist Sorry, it was a leftover from my code, mended to avoid the two-way binding silliness

mercmobily:
I am not a fan of doing a DOM query (and two of them) when you can just use the ID.
I think it's more readable without the brackets... and assigning the property feels cleaner!
I might be wrong...

H3dz:
Yes, you're right, looks like assigning the property is the way to do it.

If you look at app-drawer there's actually a "close" method that does exactly that.

`if (!this.$.drawer.persistent) this.$.drawer.close();`

mercmobily:
Absolutely agreed -- I didn't realise there was a `close()` method. Mended!

mercmobily:
(this is gonna have to be the most discussed one-liner in the history of Polymer :D )

robdodson:
Thanks @mercmobily and sorry for not getting to this sooner (i'm traveling at the moment).
Would you mind adding braces to the `if` statement? That matches the style that we use for the rest of our code.

mercmobily:
Sure!

1)
Like this:

    if (!this.$.drawer.persistent){ this.$.drawer.close(); }

Or like this:

    if (!this.$.drawer.persistent) {
      this.$.drawer.close();
    }

?
I am assuming it's the latter. Committed already -- let me know if I got it right :D
I have a newline after `this.page = page || 'view1';` as it's a separate issue logically speaking. Again, let me know if you agree.

2) Do you have a link at the style guide?

3) Do you have a beautifier that will make sure the code follows the standards?

Ta!

robdodson:
Thanks!

> 2) Do you have a link at the style guide?

No but I should come up with one.

> 3) Do you have a beautifier that will make sure the code follows the standards?

No but we should probably add one. I've [opened an issue to do so](https://github.com/PolymerElements/polymer-starter-kit/issues/933).

mercmobily:
Yay!!! My first pull request accepted into Polymer!!
Hopefully many many more to come! :D

(I am going to work on file attachments and iron-ajax soon!)

robdodson:
👍

PR #938, opened 2016-10-28
Fix paths in my-app.html

afspear:
Changing the path allowed the app to render.


afspear:
I signed it!

robdodson:
We don't generate an `app/` directory so you shouldn't need to do this. Can you explain what lead you to needing this change?

PR #940, opened 2016-10-31
Change to relative paths

chuckh:
@robdodson @abdonrd PTAL.

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,10 +17,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <title>My App</title>
     <meta name="description" content="My App description">
 
-    <link rel="icon" href="/images/favicon.ico">
+    <link rel="icon" href="images/favicon.ico">
 
     <!-- See https://goo.gl/OOhYW5 -->
-    <link rel="manifest" href="/manifest.json">
+    <link rel="manifest" href="manifest.json">
 
     <!-- See https://goo.gl/qRE0vM -->
     <meta name="theme-color" content="#3f51b5">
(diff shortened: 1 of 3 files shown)


src/my-view404.html
```diff
@@ -20,7 +20,7 @@
       }
     </style>
 
-    Oops you hit a 404. <a href="/">Head back to home.</a>
+    Oops you hit a 404. <a href="../">Head back to home.</a>
```
robdodson: Hm if someone types "foo.com/bar/baz" I don't think this would work

robdodson: I would suggest leaving this one as a slash for now and adding that comment

jsilvermist:
What about using the HTML5 base tag for using relative URLs?

robdodson:
I think/hope we should be fine with the relative URL changes Chuck has made. Happy to revisit this later if folks find that they need a `base` element.

index.html
```diff
@@ -35,14 +35,14 @@
     <meta name="apple-mobile-web-app-title" content="My App">
 
     <!-- Homescreen icons -->
-    <link rel="apple-touch-icon" href="/images/manifest/icon-48x48.png">
-    <link rel="apple-touch-icon" sizes="72x72" href="/images/manifest/icon-72x72.png">
-    <link rel="apple-touch-icon" sizes="96x96" href="/images/manifest/icon-96x96.png">
-    <link rel="apple-touch-icon" sizes="144x144" href="/images/manifest/icon-144x144.png">
-    <link rel="apple-touch-icon" sizes="192x192" href="/images/manifest/icon-192x192.png">
+    <link rel="apple-touch-icon" href="images/manifest/icon-48x48.png">
```
cherukumilli: @chuckh 
should there be a `sizes="48x48` on this line?

FluorescentHallucinogen:
Just tested with [Electron] (http://electron.atom.io) framework, doesn't work. :(

robdodson:
@FluorescentHallucinogen does it provide errors? what part doesn't work

abdonrd:
I remember @justinfagnani had recommended use absolute paths. But I can't remember why.

FluorescentHallucinogen:
@robdodson

Links and navigation drawer does not work.

You can test it yourself by run `npm i -g electron`, then `electron` commands and drag the built app from `build/unbundled` or `build/bundled` to the Electron window.

Screenshot

SheanYu:
Hi @robdodson 

Now it's works with Cordova CLI !
Except when the application starts, i have the same view as @FluorescentHallucinogen but the router is working i can navigate on the views...

Loïc.

pomber:
`build/bundled/service-worker.js` still uses absolute paths:  

```
/* eslint-disable quotes, comma-spacing */
var PrecacheConfig = [["/bower_components/webcomponentsjs/webcomponents-lite.min.js","02395895d5d08242c6ba93518a6da2c5"],["/index.html","c11c33a4a1e4e98d807ecb780291f663"], ...];
```

any ideas why?

robdodson:
I realize this was a mistake and we need to go back to absolute paths. Here's why:

Say you have the route `/view1/subview1` and `index.html` requests `src/my-app.html` as it does now. That will get translated into `/view1/src/my-app.html` which will 404. So absolute paths are critical if we want our routing to work properly. I'm going to send in a new PR to undo this change. We'll have to figure out another way to get Cordova and Electron working. Frankly those are not primary targets of ours and I'm not deeply familiar with either so if anyone in the community would like to pitch in on this it would be very helpful. It might require creating a separate template that works better for those projects.

FluorescentHallucinogen:
@robdodson It's not an option, and that's why: absolute paths breaks deployment to subdirectory (GitHub Pages, GitLab Pages, subdirectory on Firebase, etc.). I.e. absolute paths breaks not only Electron/Cordova.

robdodson:
@FluorescentHallucinogen I'm aware of this. I believe you can still deploy to subdirectories but you'll need to add in a workaround like https://github.com/PolymerElements/polymer-starter-kit/blob/v1.4.0/docs/deploy-to-github-pages.md

In theory as @jsilvermist mentioned you should be able to use a <base> tag but yesterday I discovered that Polymer rewrites anchor `href` attributes to be relative to components, and ignores base tags. This is not how vanilla HTML Imports work so I'm trying to get to the bottom of why Polymer does this. The folks who added that feature are out on vacation though :\

PR #942, opened 2016-11-01
Update webcomponent polyfill check to use ES6

stramel:
I made these separate commits for now so  I can easily revert any of them. I will squash before the merge if requested.

The following items were addressed:

- Updated webcomponent polyfill check to use ES6.
- Updated polymer to be even more lazy using the `lazyLoad: max` and `useNativeCSSProperties: true`.
- Updated the app import to be asynchronous. (Noticed this on a few Google projects)
- Updated the webcomponent polyfill check to be at the bottom of the body. (Noticed this on a few Google projects)

Happy to address any feedback! Thanks for the awesome app starting point!

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,10 +17,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <title>My App</title>
     <meta name="description" content="My App description">
 
-    <link rel="icon" href="/images/favicon.ico">
+    <link rel="icon" href="images/favicon.ico">
 
     <!-- See https://goo.gl/OOhYW5 -->
-    <link rel="manifest" href="/manifest.json">
+    <link rel="manifest" href="manifest.json">
 
     <!-- See https://goo.gl/qRE0vM -->
     <meta name="theme-color" content="#3f51b5">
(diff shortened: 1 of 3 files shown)


stramel:
I have signed the CLA with a different account previously. I have added this email to that one under and alternate email but it is still showing `cla: no`

admwx7:
:+1:

robdodson:
Because IE11 doesn't have support for arrow functions I'd like to hold off on the ES6 change. I *think* we could use const without it complaining but I need to double check (unless someone else has a PC handy).

abdonrd:
And why moving the styles and html import?

admwx7:
@robdodson this is going into the 2.0 preview which uses the ES6 class syntax for creating elements, which isn't supported in IE11 either.

robdodson:
@admwx7 ah didn't notice that, thanks :)

stramel:
@abdonrd I'm not entirely sure what the preferred placement is. I had noticed this on a couple of repositories that @notwaldorf had made like [mojibrag](https://github.com/notwaldorf/mojibrag/blob/master/index.html#L122). 

My guess is that it would give the basic styling and layout quicker. This would come in handy when adding a loader into the `index.html` page. 

I can revert that change if it is better to leave at the top.

jay8t6:
+1

robdodson:
I think, since the import is async, it should start fetching but still allow the page to move on to rendering the critical styles. This could lead to FOUC though since my-app will no longer block rendering. If you're using critical styles to start painting a shell earlier then that's fine (good, in fact) but if you're not planning for that and just carelessly putting things in index.html then the page will flicker a bit

robdodson:
I'm cool with merging this. My only request would be that you switch from using absolute paths to using relative paths. We just updated the current master of PSK to use relative paths because it helps out a bit with cordova/electron. Still working on a fixes for those folks but it seemed to be an improvement so might as well do it here

stramel:
@robdodson I have change the paths from absolute to relative paths according to the #940 PR that just got merged into master a bit ago. Would you like me to squash the commits before it is merged?

stramel:
I am getting an error when the Travis Build runs. I'm not sure how to fix this issue.

image

robdodson:
@stramel that might just be a travis/github quirk that'll sort itself out when we run it again. You don't need to squash your changes, I can do it in GitHub now

stramel:
@robdodson Anything left that I need to address?

robdodson:
sorry for the delay, got caught up in chrome dev summit. Just checking with the team to see why iron-resizable-behavior keeps failing

robdodson:
OK no clue why that line is failing. It works fine locally. Going to go ahead and merge so we're not blocked on this.

PR #951, opened 2016-12-13
Add ESLint and fixup errors

robdodson:
@abdonrd @chuckh PTAL


package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "polymer-starter-kit",
+  "license": "https://polymer.github.io/LICENSE.txt",
```
abdonrd: Change to `"SEE LICENSE IN https://polymer.github.io/LICENSE.txt"`?

Reference to [package.json#license docs](https://docs.npmjs.com/files/package.json#license).

.eslintrc
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "google",
```
abdonrd: Maybe `["eslint:recommended", "google"]`?

Reference to [Using the `google` config with `eslint:recommended`](https://github.com/google/eslint-config-google#using-the-google-config-with-eslintrecommended)

.eslintrc
```diff
@@ -0,0 +1,13 @@
...
+  "env": {
+    "browser": true
+  },
+  "plugins": [
+    "html"
+  ],
+  "rules": {
+    "no-var": "off",
```
abdonrd: Why override rules?

.eslintrc
```diff
@@ -0,0 +1,13 @@
...
+    "new-cap": "off"
```
web-padawan: Consider using `"new-cap": ["error", { "capIsNewExceptions": ["Polymer"] }]`

package.json
robdodson: Every example I could find in the Polymer and PolymerElements org uses this approach https://github.com/PolymerElements/test-fixture/blob/master/package.json#L16 so I'll switch to that. Our open source office likes for us to use the same pattern in all repos

.eslintrc
robdodson: I didn't think we needed to move to const/let just yet. I was going to hold off on that till Polymer 2 comes out

robdodson: I think we need to add `Polymer: true` because you can write to it when doing settings https://www.polymer-project.org/1.0/docs/devguide/settings

abdonrd: Oh, right! `Polymer: true`.

robdodson:
ok pushed some updates

service-worker.js
```diff
@@ -7,6 +7,9 @@
  * Code distributed by Google as part of the polymer project is also
  * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
  */
+
+/* eslint no-console: ["error", { allow: ["info"] }] */
```
abdonrd: Why not add in the `.eslintrc.json`?

test/my-view1.html
```diff
@@ -30,6 +30,8 @@
     </test-fixture>
 
     <script>
+      /* eslint-env mocha */
+      /* global fixture assert */
```
abdonrd: What do you think about to create another `.eslintrc.json` in the `test` folder and remove all the comments?

```
{
  "env": {
    "mocha": true
  },
  "globals": {
    "assert": false,
    "fixture": false,
    "WCT": false
  }
}
```

Reference: [ESLint docs: Configuration Cascading and Hierarchy](http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy)

.eslintrc
```diff
@@ -8,6 +8,9 @@
   ],
   "rules": {
     "no-var": "off",
-    "new-cap": "off"
+    "new-cap": ["error", { "capIsNewExceptions": ["Polymer"] }]
```
abdonrd: Since we have the `Polymer` global, do we need this?

service-worker.js
robdodson: In general we don't want folks leaving any `console` statements in their code. In this case we intended it.

.eslintrc
robdodson: yes otherwise it triggers an error. just tested

test/my-view1.html
robdodson: hm we could try that

robdodson:
updated

robdodson:
@abdonrd added to travis.yml. let's see if it works :D

PR #954, opened 2017-01-09
Add .gitignore as ignore-path to ESLint

abdonrd:
Add .gitignore as ignore-path to ESLint to avoid lint the `build` folder.

Also, remove the `exit 0;` to avoid the false positive.


robdodson:
thanks @abdonrd!

Issue #964, opened 2017-01-26
[ Introductory video/presentation/Screenshot needed]


vikaskyadav:
Can somebody include an introductory video or presentation or screenshots related with usage of polymer-starter-kit for beginners?

web-padawan:
@vikaskyadav there is a [polycast](https://www.youtube.com/watch?v=HgJ0XCyBwzY) about current PSK version.

vikaskyadav:
@web-padawan , then I request you to place a link on the front. It would be easy for newcomers on your repo for better understanding I guess.

web-padawan:
I'm not a maintainer of PSK, but adding that video to the readme is a good idea. @robdodson WDYT?

robdodson:
Sure

PR #965, opened 2017-01-26
Add getting started polycast to readme

robdodson:


robdodson:
Fixes https://github.com/PolymerElements/polymer-starter-kit/issues/964

PR #970, opened 2017-02-14
Ask to be linted as a hybrid app

rictic:
Part of the unreleased, configurable, new linter. More info: Polymer/polymer-cli#566

### DIFF
--- a/polymer.json
+++ b/polymer.json
@@ -15,5 +15,8 @@
   "includeDependencies": [
     "manifest.json",
     "bower_components/webcomponentsjs/webcomponents-lite.min.js"
-  ]
+  ],
+  "lint": {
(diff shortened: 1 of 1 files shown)


robdodson:
Should we wait till the linter is released to merge this?

rictic:
Your call. It causes no harm, extra properties in polymer.json will always be ignored. Storing the configuration here means that we can more easily run the new linter against psk in its integration tests.

The new linter will be merged shortly, though it will be under an alternative command while we port the rest of the lint rules from polylint, so it'll be run as something like `polymer experimental-lint`

robdodson:
could we do the change on the 2.0-preview branch of PSK instead? would your tests still be able to run against that?

rictic:
Yep, that'll work!

PR #967, opened 2017-02-06
Convert psk to use polymer 2.0

FredKSchott:
- [X] Convert to 2.0 dependencies
- [X] Convert to 2.0 Class-based syntax
- [X] Use new slots interface on app-layout
- [X] Get rid of `Polymer.dom()` wrapper where possible

---

- Merges into a new `2.0-preview` branch instead of `master`.
- Tested with `polymer serve`, `polymer lint`, `polymer build` & `polymer test`

/cc @robdodson @justinfagnani

PR #973, opened 2017-02-21
Update to latest keys

stramel:
This fixes warnings in the latest polymer-builds


stramel:
Closing since it is already in #967

PR #974, opened 2017-02-21
Fix: Add note on running WCT on Windows

stramel:


robdodson:
@rictic can you double check this?

stramel:
@robdodson need may be a bit strong of a word but without those set it takes at least a couple minutes to spin up tests versus with them set instantly running

stramel:
@robdodson @rictic Bump

PR #982, opened 2017-03-09
Fix #981 - Making the starter kit ready for static web servers

### LINKED ISSUE #981: Absolute paths on index.html
### Description

In ` index.html`  all the paths used are absolute ones, either for image resources of for the polyfill or for the main component, ` my-app` .

But in many cases, specially when doing code labs, people don't launch the server with the polymer app at its root, they deploy the app in a subfolder. Or they are simply using static webservers. And then all the links are broken.

I understand that it would also need to add `useHashAsPath` to `app-location`, but I see that like something good, as you already need to do it if using, for example a static web server.

### Expected outcome

I would like to have a `index.html` file with relative paths and hash as a stand-in alternative to path for navigation, that would work in any folder structure of the server.

### Actual outcome

Well, the current absolute path `index.html`

### Live Demo

See `index.html`

### Steps to reproduce

See `index.html`

### Browsers Affected
- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [x] Safari 8
- [x] Safari 7
- [x] Edge
- [x] IE 11
- [x] IE 10

LostInBrittany:
Relative paths in index.html and hash instead of path for navigation.

In index.html all the paths used are absolute ones, either for image resources of for the polyfill or for the main component, `my-app ` .

But in many cases, specially when doing code labs, people don't launch the server with the polymer app at its root, they deploy the app in a subfolder. Or they are simply using static webservers. And then all the links are broken.

Making paths relative also require to add `useHashAsPath` to `app-location`, but that's a nice side effect (IMHO)  as you already need to do it when deploying in static web servers.

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,10 +18,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <title>My App</title>
     <meta name="description" content="My App description">
 
-    <link rel="icon" href="/images/favicon.ico">
+    <link rel="icon" href="./images/favicon.ico">
 
     <!-- See https://goo.gl/OOhYW5 -->
-    <link rel="manifest" href="/manifest.json">
+    <link rel="manifest" href="./manifest.json">
 
     <!-- See https://goo.gl/qRE0vM -->
     <meta name="theme-color" content="#3f51b5">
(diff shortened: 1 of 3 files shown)


robdodson:
Commented https://github.com/PolymerElements/polymer-starter-kit/issues/981#issuecomment-285389927

abdonrd:
Thanks @LostInBrittany! 👏
But the #995 (already merged) fix it.

PR #992, opened 2017-03-28
fixes #991 - initial load issue on mobile

### LINKED ISSUE #991: Display 404 page for valid routes on load - 2.0-preview
### Description
Display 404 page for valid routes on load.

### Expected outcome

Display the actual page if it exists, otherwise show 404 page.

### Actual outcome

1. App shows 404 view on page load for any of the valid routes (/, /view1, /view2, /view3).
2. When navigating to a different view through the drawer it works, but on refresh, 404 is shown again.
3. There are no errors in Dev Tools, is like `Polymer.importHref` fails the first time then it works afterwards, maybe some dependencies are missing at that point.

### Steps to reproduce

1. Clone polymer-starter-kit.
2. Run firebase init and enable url rewrite in Firebase.
3. `$ polymer build --add-service-workers`
3. `$ firebase deploy`
4. Open the URL and route to a valid route.
5. Route not working, shows 404 view and no error in the Dev Tools.
6. Click on a menu item in the drawer, it should show the corresponding view.
7. Refresh the page, it show the 404 view again.

**Firebase routing:**
```json
"rewrites": [
      {
        "source": "!{/bower_components,/src}/**",
        "destination": "/index.html"
      }
    ]
```

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10
- [x] Safari on iOS 10.3
- [x] Chrome 57.0.2987.100 on iOS 10.3

davidcv5:
polymer-starter-kit is showing a 404 page when loading a valid route on the browser, though the navigation works fine once the page finished loading. Refreshing brings the 404 view again.
This PR fixes the issue.

### DIFF
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -146,13 +146,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       _pageChanged(page) {
         // Load page import on demand. Show 404 page if fails
         var resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
-        Polymer.RenderStatus.afterNextRender(this, function() {
+        Polymer.RenderStatus.afterNextRender(this, () => {
           Polymer.importHref(
               resolvedPageUrl,
               null,
               this._showPage404.bind(this),
               true);
-        }.bind(this));
+        });
       }
 
       _showPage404() {
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -146,11 +146,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       _pageChanged(page) {
         // Load page import on demand. Show 404 page if fails
         var resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
-        Polymer.importHref(
-            resolvedPageUrl,
-            null,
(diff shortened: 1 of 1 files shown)


jsilvermist:
Why use `function() {...}.bind(this)` vs `() => {...}`?

davidcv5:
@jsilvermist for some reason it failed for me before, I just tried it again and is working fine now. Thank you for the feedback.

stramel:
Feel like this should use Lazy Imports instead

abdonrd:
I can't reproduce this issue. Can you check with the last `2.0-preview` branch?

Thanks in advance!

stramel:
@abdonrd I believe I was reproducing this even in the 2.0-preview branch. I think there are a bunch of issues logged around this as well. I can take a look again here in a few minutes

abdonrd:
@stramel I just see your edit comment.

Can you confirm that the problem is no longer happening? To close the PR.

abdonrd:
Let me know if anyone can reproduce it to reopen.

Issue #919, opened 2016-09-15
How can i do work Polymer-Starter-Kit with Cordova ?


SheanYu:
Hi !

I tried to do work Polymer-Starter-Kit with the Cordova CLI but when i run it on my phone i have a blank page.

I searched on internet but i didn't found, maybe that the answer is hide too deep !

Is there someone know how to do ?

Thanks.
Greetings Loïc.

FluorescentHallucinogen:
The problem is in using absolute paths instead of relative. See #905.

robdodson:
We did a change in https://github.com/PolymerElements/polymer-starter-kit/pull/940. @SheanYu could you try the master branch and see if it helps?

robdodson:
Friendly ping, @SheanYu have you had a chance to try the new version with relative paths?

SheanYu:
Greetings @robdodson,

Yes i tried it with the last branch with the relative paths.

I answered on it, it works ! Except that when the application starts i have the error404 page with the redirection. If i follow it then it the router work Well !

Loïc.

robdodson:
ok we had to back out the relative paths change because it caused other issues https://github.com/PolymerElements/polymer-starter-kit/pull/940#issuecomment-261645920

This also helped me uncover another bug/feature https://github.com/Polymer/polymer/issues/4213

I am disinclined to make changes to PSK to support Cordova because PSK's primary goal is just to give folks an easy place to play around with Polymer. If we have to jump through hoops for Cordova then I think it would be a better strategy for someone in the community to create a Cordova generator for Polymer CLI that's based on either PSK or the Custom Build generator https://github.com/PolymerElements/generator-polymer-init-custom-build

I'm going to close this issue because I don't think we'll be switching to relative paths anytime soon and that sounds like a deal breaker for Cordova. Instead I'd recommend folks follow the linked Polymer issue from above if they want to track a workaround for this, since it's more of a general Polymer issue and not specific to PSK

PR #995, opened 2017-04-04
Change PSK 3.0 / 2.0-preview to use relative paths for subdirectory deployment

jsilvermist:
(PSK 3.0/2.0-preview) Implements relative paths as talked about on #981 without using `use-hash-as-path`.

Uses a `<base>` tag with Polymer's `rootPath`.

Here's a live preview built using `polymer build`, and using a pretty generic `.htaccess` for routing:
[PSK on non-root directory](https://silverlinkz.net/_test/polymer/2.0-relative-starter-kit/)

Basically functions the same as current PSK, but much easier to use on subdirectories by simple changing the `<base>`  tag `href`, and having the routes automatically change accordingly.

PS: This method can also be used for PSK 2.0 after `Polymer#1.9.0` lands with backport for `rootPath`.

### DIFF
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -129,7 +129,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       constructor() {
         super();
 
-        // Get root pattern for app-route
+        // Get root pattern for app-route, for more information about `rootPath` see:
+        // https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates
         this.rootPattern = (new URL(this.rootPath)).pathname;
       }
 
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -66,7 +66,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <app-location route="{{route}}"></app-location>
     <app-route
         route="{{route}}"
-        pattern="/:page"
+        pattern="[[rootPattern]]:page"
         data="{{routeData}}"
         tail="{{subroute}}"></app-route>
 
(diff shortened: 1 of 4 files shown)


src/my-app.html
```diff
@@ -126,6 +126,13 @@
         ];
       }
 
+      constructor() {
+        super();
+
+        // Get root pattern for app-route
```
robdodson: Would you mind including a link to the docs for rootPath so folks know where it's coming from?

jsilvermist: Done!

jsilvermist: Opps, abbreviated "information" to "info" on that comment to pass lint.

FredKSchott:
This is a great example of how to use a base URL to host on a subdirectory, but it also adds some complexity for everyone who is starting out trying to understand Polymer/PSK. I'll leave merging this to @robdodson since I'm not sure where this falls between those goals for the project.

robdodson:
I think we should go for it, mainly because we get _soooooo_ many issues related to folks trying to deploy to subdirectories, in particular on GH Pages.

FluorescentHallucinogen:
I'm developing a web app based on Polymer Starter Kit. Since at this time geolocation is not available in the background for PWA (Geolocation API is not available in Service Workers), I need to pack my web app into Android app using Cordova to use the native Android module (Cordova plug-in) for background geolocation.

If I use PSK v3.0.1 I've just change the absolute paths to relative in `index.html`, pack it to Android app using Cordova, run it on my Android device, and all works like a charm (except that when the app starts I have the error 404 page). Without changing the absolute paths to relative I have a blank page. See #919 for more info.

If I use PSK v3.0.2, when I pack it to Android app using Cordova and run it on my Android device, I have a blank page.

I believe the problem is in this patch. It changes the absolute paths to relative (seems, this is what is needed to fix PSK work in Cordova), but it also adds `base`, `baseUrl`, `rootPattern`, `rootPath` things.

How to fix the PSK work in Cordova after this patch was merged?

FluorescentHallucinogen:
@jsilvermist @robdodson @abdonrd @FredKSchott @justinfagnani Please help.

abdonrd:
I have not used Cordova recently.

FluorescentHallucinogen:
BTW, it affects Electron too.

abdonrd:
@FluorescentHallucinogen can you try with the actual [2.0-preview branch](https://github.com/PolymerElements/polymer-starter-kit/tree/2.0-preview)?

FluorescentHallucinogen:
Just tested with the actual `2.0-preview` branch. Still doesn't work. :(

The easiest way to test (reproduce) the problem is install Electron (type and run `npm i -g electron` command in console), run it (type and run `electron` command in console) and drag the built app (`index.html `files from `/build` subfolders) to the Electron window.

abdonrd:
I see electron serve the files as `file:///`... Maybe you need to create a server?

Sorry, I never used Electron.

PR #997, opened 2017-04-07
New: Utilize lazy imports

stramel:


bower.json
```diff
@@ -13,6 +13,7 @@
     "iron-media-query": "PolymerElements/iron-media-query#^1.0.0",
     "iron-pages": "PolymerElements/iron-pages#^1.0.0",
     "iron-selector": "PolymerElements/iron-selector#^1.0.0",
+    "lazy-imports": "Polymer/lazy-imports#master",
```
stramel: NOTE: I got @garlicnation to merge some changes but he forgot to tag them. :sweat_smile:

src/my-app.html
```diff
@@ -126,8 +136,12 @@
 
       _pageChanged: function(page) {
         // Load page import on demand. Show 404 page if fails
-        var resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
-        this.importHref(resolvedPageUrl, null, this._showPage404, true);
+        this.importLazyGroup(page).
+          then(function (results) {
+            if (results.failed.length > 0 || results.loaded.length < 1) {
```
stramel: I'm hoping we can get this cleaned up to use catch instead.

robdodson:
I think we'd prefer to stick with regular HTML Imports as PSK is meant as a starting point for new users. If you want a more feature rich app boilerplate it might make sense to create a separate generator

stramel:
@robdodson This was proposed in the last Polymer Summit by @justinfagnani. It will help with static analysis and prevent the warnings of using an element that doesn't have an import. It seems like it would be easier for new users to use as they don't have to name their components my-*.html and can structure it easily as well.

robdodson:
I'll leave it up to @justinfagnani. I see the value but also have mixed feelings about introducing yet another thing to the new user's experience. It doesn't seem like lazy-imports has seen a recent release so I'd at least like to be able to depend on a version instead of `#master`. Also, PSK is tied in to our tutorial so any changes to it need to be reflected in the docs (https://www.polymer-project.org/2.0/start/toolbox/set-up).

stramel:
I understand the mixed feelings. I thought it would make it easier to understand since you would have the import still. 

I would like to get it tagged, but @garlicnation forgot to last time he was on. 

I would be more than happy to add documentation for it on both 1.0 and 2.0 if you guys approve this direction.

abdonrd:
If the `Polymer/lazy-imports` becomes stable and official (with polymer tools support: analyze, lint, build, etc), I like the idea to add it to PSK.

stramel:
@abdonrd @robdodson 

Just wanted to update you guys with the latest status:

- I have followed up on a bunch of issues and PRs in the Polymer Tooling to make sure lazy-imports works. You can see a semi-complete list [here](https://github.com/Polymer/lazy-imports/issues/23) Thanks @usergenic 
- I have a PR out to do some clean up, waiting for someone to take a look (https://github.com/Polymer/lazy-imports/pull/25)
- Still waiting for someone to tag a version before the `2.0-preview` got merged in.
- Since `2.0-preview` got merged in, you have to add resolutions on `polymer` and `webcomponents` until they are tagged.
- On the bright side, it is a Hybrid component.

So it looks like we are getting close :)

abdonrd:
Maybe we should use the lazy-import only in the `PSK#2.0-preview`.
What is your opinion? @stramel @FredKSchott

stramel:
@abdonrd I would be okay with that. I'm hoping I can get a branch of `lazy-imports` for `1.x` to address one issue and we could point to that if we decide to use them here.

stramel:
@abdonrd @robdodson @justinfagnani @FredKSchott 

`lazy-imports` has been tagged at `v2.0.0`.

Still waiting for thoughts on using it in the PSK or not.

stramel:
With the merge of #1016 I'm going to close this in favor of #998

PR #998, opened 2017-04-07
New: Implement lazy-imports

stramel:
**TODO:**

- [x] Merge PRs in `lazy-imports`
    
- [ ] Add documentation to Polymer Docs and Tooling
    - [ ] Polymer Docs https://github.com/Polymer/docs/pull/2279 (Thanks @arthurevans and @katejeffreys)
    - [ ] Polymer Tooling (Hopefully someone could help me out here?)
- [x] Profiling of `lazy-imports`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,17 +1,15 @@
 {
   "extends": ["eslint:recommended", "google"],
-  "parserOptions": {
-    "ecmaVersion": 6
-  },
   "env": {
-    "browser": true
+    "browser": true,
+    "es6": true
   },
   "plugins": [
     "html"
   ],
   "rules": {
     "brace-style": "off",
-    "new-cap": ["error", { "capIsNewExceptions": ["Polymer"] }],
+    "new-cap": ["error", { "capIsNewExceptions": ["Polymer"], "capIsNewExceptionPattern": "Mixin$" }],
     "no-var": "off",
     "require-jsdoc": "off"
   },
--- a/polymer.json
+++ b/polymer.json
@@ -1,12 +1,6 @@
 {
   "entrypoint": "index.html",
   "shell": "src/my-app.html",
-  "fragments": [
-    "src/my-view1.html",
-    "src/my-view2.html",
-    "src/my-view3.html",
-    "src/my-view404.html"
-  ],
   "sources": [
     "images/**/*",
     "src/**/*"
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -121,6 +121,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // preventable, allowing for better scrolling performance.
     Polymer.setPassiveTouchGestures(true);
 
+    /**
+     * @extends Polymer.Element
+     * @appliesMixin Polymer.LazyImportsMixin
+     */
     class MyApp extends Polymer.LazyImportsMixin(Polymer.Element) {
       static get is() { return 'my-app'; }
 
@@ -159,10 +163,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       _pageChanged(page) {
         // Load page import on demand. Show 404 page if fails
         this.importLazyGroup(page).
-          then((results) => {
-            if (results.failed.length > 0 || results.loaded.length < 1)
-              this._showPage404();
-          });
+          catch(this._showPage404.bind(this));
       }
 
       _showPage404() {
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -121,10 +121,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // preventable, allowing for better scrolling performance.
     Polymer.setPassiveTouchGestures(true);
 
-    /**
-     * @extends Polymer.Element
-     * @appliesMixin Polymer.LazyImportsMixin
-     */
     class MyApp extends Polymer.LazyImportsMixin(Polymer.Element) {
       static get is() { return 'my-app'; }
 
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -19,15 +19,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="../bower_components/app-route/app-route.html">
 <link rel="import" href="../bower_components/iron-pages/iron-pages.html">
 <link rel="import" href="../bower_components/iron-selector/iron-selector.html">
+<link rel="import" href="../bower_components/lazy-imports/lazy-imports-mixin.html">
 <link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="my-icons.html">
 
-<link rel="lazy-import" href="my-view1.html">
-<link rel="lazy-import" href="my-view2.html">
-<link rel="lazy-import" href="my-view3.html">
-<link rel="lazy-import" href="my-view404.html">
-
 <dom-module id="my-app">
+  <link rel="lazy-import" group="view1" href="my-view1.html">
+  <link rel="lazy-import" group="view2" href="my-view2.html">
+  <link rel="lazy-import" group="view3" href="my-view3.html">
+  <link rel="lazy-import" group="view404" href="my-view404.html">
+
   <template>
     <style>
       :host {
@@ -120,7 +121,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // preventable, allowing for better scrolling performance.
     Polymer.setPassiveTouchGestures(true);
 
-    class MyApp extends Polymer.Element {
+    class MyApp extends Polymer.LazyImportsMixin(Polymer.Element) {
       static get is() { return 'my-app'; }
 
       static get properties() {
@@ -157,12 +158,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _pageChanged(page) {
         // Load page import on demand. Show 404 page if fails
-        const resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
-        Polymer.importHref(
-            resolvedPageUrl,
-            null,
-            this._showPage404.bind(this),
-            true);
+        this.importLazyGroup(page).
+          then((results) => {
+            if (results.failed.length > 0 || results.loaded.length < 1)
+              this._showPage404();
+          });
       }
 
       _showPage404() {
--- a/bower.json
+++ b/bower.json
@@ -12,7 +12,7 @@
     "iron-media-query": "PolymerElements/iron-media-query#^2.0.0",
     "iron-pages": "PolymerElements/iron-pages#^2.0.0",
     "iron-selector": "PolymerElements/iron-selector#^2.0.0",
-    "lazy-imports": "Polymer/lazy-imports#^2.0.0",
+    "lazy-imports": "Polymer/lazy-imports#^2.1.0",
     "paper-icon-button": "PolymerElements/paper-icon-button#^2.0.0",
     "polymer": "Polymer/polymer#^2.0.0",
     "webcomponentsjs": "webcomponents/webcomponentsjs#^1.0.0"
--- a/bower.json
+++ b/bower.json
@@ -12,6 +12,7 @@
     "iron-media-query": "PolymerElements/iron-media-query#^2.0.0",
     "iron-pages": "PolymerElements/iron-pages#^2.0.0",
     "iron-selector": "PolymerElements/iron-selector#^2.0.0",
+    "lazy-imports": "Polymer/lazy-imports#^2.0.0",
     "paper-icon-button": "PolymerElements/paper-icon-button#^2.0.0",
     "polymer": "Polymer/polymer#^2.0.0",
     "webcomponentsjs": "webcomponents/webcomponentsjs#^1.0.0"
--- a/src/my-view1.html
+++ b/src/my-view1.html
@@ -30,9 +30,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   </template>
 
   <script>
-    /**
-     * @extends Polymer.Element
-     */
     class MyView1 extends Polymer.Element {
       static get is() { return 'my-view1'; }
     }
--- a/src/my-view1.html
+++ b/src/my-view1.html
@@ -30,6 +30,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   </template>
 
   <script>
+    /**
+     * @extends Polymer.Element
+     */
     class MyView1 extends Polymer.Element {
       static get is() { return 'my-view1'; }
     }
--- a/src/my-view2.html
+++ b/src/my-view2.html
@@ -30,9 +30,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   </template>
 
   <script>
-    /**
-     * @extends Polymer.Element
-     */
     class MyView2 extends Polymer.Element {
       static get is() { return 'my-view2'; }
     }
--- a/src/my-view2.html
+++ b/src/my-view2.html
@@ -30,6 +30,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   </template>
 
   <script>
+    /**
+     * @extends Polymer.Element
+     */
     class MyView2 extends Polymer.Element {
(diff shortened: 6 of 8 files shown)


src/my-app.html
```diff
@@ -19,10 +19,16 @@
 <link rel="import" href="../bower_components/app-route/app-route.html">
 <link rel="import" href="../bower_components/iron-pages/iron-pages.html">
 <link rel="import" href="../bower_components/iron-selector/iron-selector.html">
+<link rel="import" href="../bower_components/lazy-imports/lazy-imports-mixin.html">
 <link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="my-icons.html">
 
 <dom-module id="my-app">
+  <link rel="lazy-import" group="view1" href="src/my-view1.html">
```
stramel: I think I'm missing something in the `lazy-import`'s 2.0-preview in the `Polymer.ResoveUrl.resolveUrl` method with the `baseUri`

stramel:
Also, getting `HTMLImports.whenReady is not a function` error when it loads the lazy imports but it isn't blocking the imports.

stramel:
Pending decision in #997

index.html
```diff
@@ -11,6 +11,7 @@
 <!doctype html>
 <html lang="en">
   <head>
+    <base href="/">
```
web-padawan: This would be a pain for those asking in Slack "how to serve PSK from a subfolder?"

stramel: I can remove this. Not entirely sure why I added it, sorry!

stramel:
Looks like eslint is yelling at the capital on the mixin... thought it was correct.

jsilvermist:
Seems to me more of an eslint issue than a naming issue. After the rebase there are lazy imports both inside and outside of the dom-module, is this intentional? Also, what was the intention of the lazy imports outside the dom-module?

stramel:
@jsilvermist :scream: WUT... I don't even know... Fixing Now. Good catch!! Lazy-Imports should be in the dom-module.

I guess, wrt eslint, we should figure out how we want to go about this. I'm fully open to suggestions :)

stramel:
@jsilvermist I added a regex `Mixin$` for exceptions to the `newCap` rule. Again, open to suggestions!

stramel:
Rebased it, so it is a bit cleaner... 

@jsilvermist Seems like those other lazy-imports accidentally got in from @justinfagnani https://github.com/PolymerElements/polymer-starter-kit/commit/f037cfb00783a6cdd81d3728a8a0b1cee301f077

.eslintrc.json
```diff
@@ -1,17 +1,15 @@
 {
   "extends": ["eslint:recommended", "google"],
-  "parserOptions": {
-    "ecmaVersion": 6
-  },
   "env": {
-    "browser": true
+    "browser": true,
+    "es6": true
```
stramel: `parserOptions` is set by this line.

http://eslint.org/docs/user-guide/configuring#specifying-environments

src/my-app.html
```diff
@@ -161,12 +162,11 @@
...
         // Load page import on demand. Show 404 page if fails
-        var resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
-        Polymer.importHref(
-            resolvedPageUrl,
-            null,
-            this._showPage404.bind(this),
-            true);
+        this.importLazyGroup(page).
+          then((results) => {
+            if (results.failed.length > 0 || results.loaded.length < 1)
```
abdonrd: It is the best option?

cc @justinfagnani

src/my-app.html
```diff
@@ -111,7 +112,7 @@
   </template>
 
   <script>
-    class MyApp extends Polymer.Element {
+    class MyApp extends Polymer.LazyImportsMixin(Polymer.Element) {
```
abdonrd: We should add the annotations here?

```
/**
 * @extends Polymer.Element
 * @appliesMixin Polymer.LazyImportsMixin
 */
```

cc @rictic

.eslintrc.json
```diff
@@ -1,17 +1,15 @@
...
   },
   "plugins": [
     "html"
   ],
   "rules": {
     "brace-style": "off",
-    "new-cap": ["error", { "capIsNewExceptions": ["Polymer"] }],
+    "new-cap": ["error", { "capIsNewExceptions": ["Polymer"], "capIsNewExceptionPattern": "Mixin$" }],
```
abdonrd: I think we can remove this rule.

abdonrd (changes requested):

.eslintrc.json
stramel: @abdonrd Removing this rule results in eslint error on `LazyImportsMixin`.

src/my-app.html
stramel: /sub

Curious about this 👍

src/my-app.html
stramel: Not sure how else we would detect if a page has loaded or not to fall back to a 404 page.

.eslintrc.json
abdonrd: Oh, right, is enable by eslint-config-google!

src/my-app.html
stramel: Added some based off of this PR: https://github.com/Polymer/polymer/pull/4568/

graynorton:
Per the discussion in https://github.com/Polymer/polymer-cli/issues/748, I think we'll want to remove the explicit declaration of fragments in `polymer.json` as part of this change.

stramel:
@graynorton Removed them after discussion with @usergenic  and some initial testing, it appears that it will treat lazy-imports as implicit fragments since they are included in a fragment, the app-shell in this case.

polymer.json
```diff
@@ -1,12 +1,7 @@
 {
   "entrypoint": "index.html",
   "shell": "src/my-app.html",
-  "fragments": [
-    "src/my-view1.html",
-    "src/my-view2.html",
-    "src/my-view3.html",
-    "src/my-view404.html"
-  ],
+  "fragments": [],
```
danizord: I think you can remove this `fragments` key completely as seems like it is not usable anymore with lazy-imports.

abdonrd: I think the same.

src/my-app.html
```diff
@@ -111,7 +112,13 @@
   </template>
 
   <script>
-    class MyApp extends Polymer.Element {
+    /**
+     * @customElement
+     * @polymer
+     * @extends Polymer.Element
+     * @appliesMixin Polymer.LazyImportsMixin
+     */
```
abdonrd: We should add all these annotations?

cc @rictic @justinfagnani

src/my-app.html
abdonrd: Can we get the list of the `lazy-imports` groups?

If we have an array of the groups names, we could avoid try to import a group that we already know does not exist.

stramel: @abdonrd I suppose I could query the list of lazy-imports then check the group exists?

Seems like duplicate effort when if it doesn't find that group in the same query it just fails out by returning empty arrays for both failed and loaded.

stramel:
Rebased onto the latest of 2.0-preview

src/my-app.html
abdonrd: I just saw how `Polymer.__importLazyGroup` works. You are right!

stramel:
So I think we are just waiting for @justinfagnani's review

src/my-app.html
abdonrd: Meanwhile, @stramel can you add the annotations to the other elements? To be consistent. Thanks in advance!

rictic: Yep, those annotations look correct to me. In the future we'd like to be able to infer as many of them as possible. e.g. we can probably already infer `@customElement` here because there's a `customElements.define` call for the element in this same file.

The simplest way to tell if the annotations are working is to look at the output of `polymer analyze`. @aomarks is actively working on the documentation viewer, but in the meantime you could just open the output in your favorite editor and grep around.

stramel: @abdonrd I have updated the annotations after discussing with @rictic.

`@customElement` and `@polymer` are inferred so I went ahead and dropped them.

abdonrd: Thanks to both!

arthurevans:
@justinfagnani hasn't weighed in, so I'm going to jump in and say we shouldn't merge this into PSK until we've done a good deal more groundwork to make sure lazy-imports are a thing.

Currently, lazy-imports aren't documented on the polymer site, or in the READMEs for polymer-cli or polymer-build. Even the lazy-imports README doesn't describe the benefits (https://github.com/Polymer/lazy-imports/issues/11). If we add them to the default starting point for new users without documenting them, it will confuse the heck out of new users. (Or even not-so-new users like me.)

As far as I can tell, this is sugar that lets you leave your fragments out of `polymer.json`, at the cost of an extra dependency. Is that an improvement on the current system? Have we profiled it to make sure it's not a regression compared to the current system?

I know @stramel has done a bunch of work on lazy-imports, I just think we should be purposeful about this change.

stramel:
@arthurevans First off I really appreciate your input and insight.

I definitely agree that it will need better documentation in general. I can work on adding that to the Polymer Docs and Tooling (CLI, Build, Bundler, Linter, etc) although I'm pretty bad at writing so I'm sure someone will have input on refining it. Hopefully @justinfagnani can clear up the benefits of `lazy-imports` in the README.

The largest benefit of it is that it offers a declarative way to handle a lazy dependency instead of the more programmatic way of handling it (`this.importHref('element.htm', null, null, true);`). This allows the tools to do better static analysis and provide a way to closer mimic html imports. The removal of `fragments` is a nice outcome of @usergenic's awesome work on the tools (Maybe @usergenic would like to weigh in as well).

WRT, profiling and improvement, I haven't actually looked into this too much. I'm not sure how to go about profiling but I can also try looking into that a bit more.

---

On another note, I feel like the fake lazy-imports that have been added to the PSK are going to really be confusing since they are not in the proper location and actually don't do anything other than prevent the linter from throwing warnings.

arthurevans:
Thanks @stramel. What are the fake lazy-imports? I haven't seen those.

stramel:
@arthurevans These are the fake lazy-imports that I am referring to. https://github.com/PolymerElements/polymer-starter-kit/blob/2.0-preview/src/my-app.html#L25-L28

They aren't properly inside the dom-module and they aren't actually used by anything in the app.

arthurevans:
Ah, right. I was looking in master. Yeah, agree they shouldn't be there if they don't do anything.

rictic:
@arthurevans You're 100% right. We need to document lazy-imports, and I think right now is a good time to do it. I wrote up https://github.com/Polymer/lazy-imports/pull/31 as a first step

abdonrd:
@arthurevans @stramel @rictic thanks for the feedback!

About the fake `lazy-imports`... I think @justinfagnani had added it only to clean Polymer Linter output.

rictic:
The lazy-imports repo has now been documented.

stramel:
Just doing a quick profile on use of `importHref` vs `__importLazyGroup`.

Used PSK `2.0-preview` and `ms/2.0-lazy-imports` branches
Ran DevTools JavaScript Profiler.
Cleared out all application data before each run.
Loaded the main page, started profiler.
Loaded non-loaded pages and attempted re-loading previously loaded imports multiple times.

|Method                      |1   |2    |3  |4   |5  |AVG|
|:----------------------------|:--:|:---:|:--:|:---:|:--:|:----:|
|`importHref`               |5.9|3.6|5.1|4.6|4.2|4.68|
|`__importLazyGroup`|6.1|2.8|4.9|4.4|4.2|4.48|

Any help with how to better profile will be greatly appreciated. I haven't really done it before in case you couldn't tell :sweat_smile:

stramel:
I setup a rough repository for profiling a different way. https://github.com/stramel/profile-lazy-imports

Both averaged around .5 -.7 for importing 3 different assets. Lazy-imports added ~600 bytes minified from what I could tell. Again, super unofficial lol

src/my-app.html
arthurevans: It looks like we're adding a new dependency on `Promise` here in order to add lazy imports. Doesn't that mean that we need to add a `Promise` polyfill to PSK to maintain compatibility?

stramel: Good call @arthurevans! :+1: I put out a PR with a note on the readme and added `promise-polyfill` as a devDependency.

FluorescentHallucinogen:
Any progress?

stramel:
@FluorescentHallucinogen Waiting for approval on the pending PRs currently.

stramel:
The PRs for `lazy-imports` functionality have been merged. Tasks still outstanding:

- Tag v2.1.0 and v3.0.0 on `lazy-imports`
- Update the docs PR to reflect the v3.0.0 syntax
- Update this PR to reflect the v3.0.0 syntax
- Docs around tooling with `lazy-imports`?
- Profiling `lazy-imports`?

stramel:
@justinfagnani All changes to `lazy-imports` have been updated and tagged. Is there anything in here you would like to see changed? I would really like to get this merged before work on P3 gets started.

keanulee:
Given that we are moving to JS modules with Polymer 3 instead of HTML imports, lazy-imports doesn't seem like a worthwhile move. I'll leave it up to other maintainers, but I'm inclined to close this PR.

stramel:
@keanulee I figured it would be worth adding for the Polymer 2.0 version as the way to handle lazy imports. Moving to Polymer 3, I agree that `lazy-imports` is no longer necessary. Although, I don't suspect everyone will instantly shift to using Polymer 3 in the enterprise world.

FluorescentHallucinogen:
Is `lazy-imports` production-ready? If yes, we should add it to PSK and it should be last Polymer 2.x PSK release before Polymer 3.0 PSK release.

andrewspy:
I have been using `lazy-imports` for Polymer 2.0 for a while now, and I do think it should be added to the last Polymer 2.x PSK release as there are many Polymer 2.x in production now, and it may take a while for Polymer 3.x to gain traction.

FluorescentHallucinogen:
Are there any production projects that use `lazy-imports`?

BTW, what about adding `lazy-imports` to https://github.com/Polymer/shop and https://github.com/Polymer/news too?

src/my-app.html
```diff
@@ -120,7 +121,11 @@
     // preventable, allowing for better scrolling performance.
     Polymer.setPassiveTouchGestures(true);
 
-    class MyApp extends Polymer.Element {
+    /**
+     * @extends Polymer.Element
+     * @appliesMixin Polymer.LazyImportsMixin
+     */
```
abdonrd: The annotations of this PR are no longer necessary. Please, delete them.

abdonrd (changes requested):

abdonrd:
@keanulee @justinfagnani @stramel With this PR, we drop the performance Lighthouse score from 95 to 90.

#### `#master` branch
screen shot 2018-01-17 at 14 36 17

#### `#master` branch + lazy-imports PR
screen shot 2018-01-17 at 14 36 41

If this can not be improved, I am inclined not to merge it.

stramel:
We have been running `lazy-imports` at the company I work at for a while. We have launched numerous production apps that currently use `lazy-imports`. I would consider it "production-ready". 

As for performance, I did observe the same thing as @abdonrd when running lighthouse on the PSK. However, just testing manually, I have never seen much difference performance-wise between `lazy-imports` and using `importHref`. The production apps that we have at one point when I released them were getting mid to upper 90s on lighthouse. (Not sure where they are now)

@FluorescentHallucinogen I had planned on trying to get this PR through before attempting to get shop and news to use `lazy-imports`.

stramel:
@justinfagnani I will take a look soon hopefully.

stramel:
Here are the results that I found. Please let me know if I can improve something, still new to using Performance tab and profiling in general. 

https://gist.github.com/stramel/82509631e3e187760d4eb916f1ec787a

stramel:
After some further discussion over on the lazy-imports PR, I believe we came to the conclusion to not use `async` for dynamically added imports. 

@justinfagnani Seeing as `async` is the primary difference in performance here, is there anything else that is holding this PR up?

stramel:
@justinfagnani Ping

abdonrd:
@justinfagnani old PR; closing? merging? ...?

stramel:
Man, I wish we could have got this stuff in for 2.0

PR #1002, opened 2017-04-17
Update the Service Worker register comment

abdonrd:


FredKSchott:
Rebase once you drop node v4 support to fix tests.
This looks fine, is there a specific reason you're making this change?

abdonrd:
@FredKSchott nothing super special. Before the #995 we only have the SW register in the `<script>` tag. And now we also have the `baseUrl`.

PR #1008, opened 2017-04-25
Use relative paths with Polymer v1 Starter Kit

jsilvermist:
#### Polymer v1 / PSK 2.0 version of #995, basically the same with a few differences for Polymer v1 support.
1. Since `created` is called too early, and `ready` too late, ~~uses an observer to set `rootPattern`.~~ **Now uses a computed property**
1. Since Polymer v1 rewrites paths based on import path it requires `rootPath` bindings for `view` links. (Also works in Polymer 2, but not necessary.) **`rootPath` bindings will also be added to `2.0-preview`**
1. ~~Update Polymer dependency to `1.9.0` for `rootPath` support.~~ **Already done on master**

Due to the differences in rewriting of paths I'm not sure where this fits in continuity between PSK 2.0 and PSK 3.0, ease of use for new users, and ease of use when deploying to a subdirectory.

#### Live Demos
[Simple Build](https://silverlinkz.net/_test/polymer/relative-starter-kit/)
[Bundled & Minified Build](https://silverlinkz.net/_test/polymer/relative-starter-kit-bundled/)

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,14 +19,15 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <meta name="description" content="My App description">
 
     <!--
-      When deploying to a non-root path, transform the base href by adding an
-      absolute reference to the root path your application will be deployed to, by either:
+      The `<base>` tag below is present to support two advanced deployment options:
+      1) Differential serving. 2) Serving from a non-root path.
 
-      - Adding a `basePath` property to your build configuration in `polymer.json`.
-        For example: { "preset": "es5-bundled", "basePath": "/subfolder/app-root/" }
+      Instead of manually editing the `<base>` tag yourself, you should generally either:
+      a) Add a `basePath` property to the build configuration in your `polymer.json`.
+      b) Use the `--base-path` command-line option for `polymer build`.
 
-      - Using the `--base-path` command line flag for polymer build in the CLI.
-        For example: polymer build --base-path /subfolder/app-root/
+      Note: If you intend to serve from a non-root path, you will also need to manually
+      comment out one line of JavaScript, see [polymer-root-path] below.
     -->
     <base href="/">
 
@@ -65,6 +66,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         dom: 'shadow',
         lazyRegister: true,
       };
+
+      // [polymer-root-path] By default, we set `Polymer.rootPath` to the server root path
+      // to support differential serving; this setting is also compatible with standard,
+      // single-build deployments. If you intend to serve your app from a non-root path,
+      // comment out the line below.
+      window.Polymer.rootPath = window.location.origin;
 
       // Load webcomponentsjs polyfill if browser does not support native
       // Web Components
--- a/index.html
+++ b/index.html
@@ -18,10 +18,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <title>My App</title>
     <meta name="description" content="My App description">
 
-    <link rel="icon" href="/images/favicon.ico">
+    <!--
+      If deploying to a non-root path, replace href="/" with the full path to the project root.
+      For example: href="/polymer-starter-kit/relative-path-example/"
+    -->
+    <base href="/">
+
+    <link rel="icon" href="images/favicon.ico">
 
     <!-- See https://goo.gl/OOhYW5 -->
-    <link rel="manifest" href="/manifest.json">
+    <link rel="manifest" href="manifest.json">
 
     <!-- See https://goo.gl/qRE0vM -->
     <meta name="theme-color" content="#3f51b5">
@@ -36,14 +42,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <meta name="apple-mobile-web-app-title" content="My App">
 
     <!-- Homescreen icons -->
-    <link rel="apple-touch-icon" href="/images/manifest/icon-48x48.png">
-    <link rel="apple-touch-icon" sizes="72x72" href="/images/manifest/icon-72x72.png">
-    <link rel="apple-touch-icon" sizes="96x96" href="/images/manifest/icon-96x96.png">
-    <link rel="apple-touch-icon" sizes="144x144" href="/images/manifest/icon-144x144.png">
-    <link rel="apple-touch-icon" sizes="192x192" href="/images/manifest/icon-192x192.png">
+    <link rel="apple-touch-icon" href="images/manifest/icon-48x48.png">
+    <link rel="apple-touch-icon" sizes="72x72" href="images/manifest/icon-72x72.png">
+    <link rel="apple-touch-icon" sizes="96x96" href="images/manifest/icon-96x96.png">
+    <link rel="apple-touch-icon" sizes="144x144" href="images/manifest/icon-144x144.png">
+    <link rel="apple-touch-icon" sizes="192x192" href="images/manifest/icon-192x192.png">
 
     <!-- Tile icon for Windows 8 (144x144 + tile color) -->
-    <meta name="msapplication-TileImage" content="/images/manifest/icon-144x144.png">
+    <meta name="msapplication-TileImage" content="images/manifest/icon-144x144.png">
     <meta name="msapplication-TileColor" content="#3f51b5">
     <meta name="msapplication-tap-highlight" content="no">
 
@@ -53,6 +59,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         dom: 'shadow',
         lazyRegister: true,
       };
+
+      var baseUrl = document.querySelector('base').href;
 
       // Load webcomponentsjs polyfill if browser does not support native
       // Web Components
@@ -78,7 +86,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         if (!webComponentsSupported) {
           var script = document.createElement('script');
           script.async = true;
-          script.src = '/bower_components/webcomponentsjs/webcomponents-lite.min.js';
+          script.src = baseUrl + 'bower_components/webcomponentsjs/webcomponents-lite.min.js';
           script.onload = onload;
           document.head.appendChild(script);
         } else {
@@ -89,12 +97,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // Load pre-caching Service Worker
       if ('serviceWorker' in navigator) {
         window.addEventListener('load', function() {
-          navigator.serviceWorker.register('/service-worker.js');
+          navigator.serviceWorker.register(baseUrl + 'service-worker.js');
         });
       }
     </script>
 
-    <link rel="import" href="/src/my-app.html">
+    <link rel="import" href="src/my-app.html">
 
     <style>
       body {
--- a/index.html
+++ b/index.html
@@ -19,8 +19,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <meta name="description" content="My App description">
 
     <!--
-      If deploying to a non-root path, replace href="/" with the full path to the project root.
+      If deploying to a non-root path, replace href="/" with an
+      absolute reference to the root path your application will deploy to.
       For example: href="/polymer-starter-kit/relative-path-example/"
+
+      Note: When serving locally with `polymer serve` this must be set to href="/".
     -->
     <base href="/">
 
--- a/index.html
+++ b/index.html
@@ -99,7 +99,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // Load pre-caching Service Worker
       if ('serviceWorker' in navigator) {
         window.addEventListener('load', function() {
-          navigator.serviceWorker.register('service-worker.js');
+          navigator.serviceWorker.register('service-worker.js', {scope: Polymer.rootPath});
         });
       }
     </script>
--- a/index.html
+++ b/index.html
@@ -102,7 +102,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // Load pre-caching Service Worker
       if ('serviceWorker' in navigator) {
         window.addEventListener('load', function() {
-          navigator.serviceWorker.register('service-worker.js', {scope: Polymer.rootPath});
+          navigator.serviceWorker.register('service-worker.js', {
+            scope: Polymer.rootPath,
+          });
         });
       }
     </script>
--- a/index.html
+++ b/index.html
@@ -87,7 +87,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         if (!webComponentsSupported) {
           var script = document.createElement('script');
           script.async = true;
-          script.src = baseUrl + 'bower_components/webcomponentsjs/webcomponents-lite.min.js';
+          script.src = baseUrl +
+            'bower_components/webcomponentsjs/webcomponents-lite.min.js';
           script.onload = onload;
           document.head.appendChild(script);
         } else {
--- a/index.html
+++ b/index.html
@@ -67,10 +67,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         lazyRegister: true,
       };
 
-      // [polymer-root-path] By default, we set `Polymer.rootPath` to the server root path
-      // to support differential serving; this setting is also compatible with standard,
-      // single-build deployments. If you intend to serve your app from a non-root path,
-      // comment out the line below.
+      // [polymer-root-path] By default, we set `Polymer.rootPath` to the
+      // server root path to support differential serving; this setting is also
+      // compatible with standard, single-build deployments. If you intend to
+      // serve your app from a non-root path, comment out the line below.
       window.Polymer.rootPath = window.location.origin;
 
       // Load webcomponentsjs polyfill if browser does not support native
--- a/index.html
+++ b/index.html
@@ -60,6 +60,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         lazyRegister: true,
       };
 
+      // Register the base URL
       var baseUrl = document.querySelector('base').href;
 
       // Load webcomponentsjs polyfill if browser does not support native
--- a/index.html
+++ b/index.html
@@ -19,11 +19,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <meta name="description" content="My App description">
 
     <!--
-      If deploying to a non-root path, replace href="/" with an
-      absolute reference to the root path your application will deploy to.
-      For example: href="/polymer-starter-kit/relative-path-example/"
+      When deploying to a non-root path, transform the base href by adding an
+      absolute reference to the root path your application will be deployed to, by either:
 
-      Note: When serving locally with `polymer serve` this must be set to href="/".
+      - Adding a `basePath` property to your build configuration in `polymer.json`.
+        For example: { "preset": "es5-bundled", "basePath": "/subfolder/app-root/" }
+
+      - Using the `--base-path` command line flag for polymer build in the CLI.
+        For example: polymer build --base-path /subfolder/app-root/
     -->
     <base href="/">
 
--- a/index.html
+++ b/index.html
@@ -63,9 +63,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         lazyRegister: true,
       };
 
-      // Register the base URL
-      var baseUrl = document.querySelector('base').href;
-
       // Load webcomponentsjs polyfill if browser does not support native
       // Web Components
       (function() {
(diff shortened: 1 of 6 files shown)


src/my-app.html
```diff
@@ -113,9 +113,16 @@
       },
 
       observers: [
+        '_setRootPattern(rootPath)',
```
abdonrd: What do you think about change the observer to a computed property?

```javascript
rootPattern: {
  type: String,
  computed: '_computeRootPattern(rootPath)'
}
```

abdonrd:
Thanks for the PR, @jsilvermist! 👏

src/my-app.html
jsilvermist: Ah didn't think of that, I like how it gives an example of how to use computed properties right in the starter kit though! Changed!

src/my-app.html
```diff
@@ -105,6 +105,10 @@
       is: 'my-app',
 
       properties: {
+        rootPattern: {
+          type: String,
+          computed: '_computeRootPattern(rootPath)',
+        },
```
abdonrd: Nit: One line break please.

jsilvermist: Done.

abdonrd: Thanks @jsilvermist! :)

index.html
```diff
@@ -18,10 +18,16 @@
     <title>My App</title>
     <meta name="description" content="My App description">
 
-    <link rel="icon" href="/images/favicon.ico">
+    <!--
+      If deploying to a non-root path, replace href="/" with the full path to the project root.
```
FredKSchott: > with the full path to the project root.

To me this sounds like some file path thing. How about: `replace href="/" with an absolute reference to the root path your application will deploy to.`

Also, will need a note that this could not be served locally by any of our tools.

jsilvermist: Changed and added note.

As for `polymer serve`, wouldn't it be a relatively easy fix to check for a base tag in the entrypoint, (`<base[^>]*href=\"(.+?)\"`) and serve from that url? (OR|AND) Maybe with something line a -B flag to manually rewrite to a custom base url?

index.html
```diff
@@ -54,6 +63,9 @@
         lazyRegister: true,
       };
 
+      // Register the base URL
+      var baseUrl = document.querySelector('base').href;
```
graynorton: Is this actually necessary? I believe that this and the string concatenations below can be omitted, and you can just use relative paths when setting `src` in your dynamically created `<script>` elements; the browser will take the document's base URL into account when fetching the source for these elements.

graynorton: BTW, we are working on support for multi-build deployments with differential serving, which would rely on this `<base>` tag scheme, so I'd like to get this merged and picked up by the CLI sooner, rather than later.

jsilvermist: Yes, it would normally work without the `base` tag, except when I tested with nested paths, eg.
Starter kit at `/psk/basedir/`
and accessing `/psk/basedir/view/subview/`
Without the base tag and concats it was looking for the files in `/psk/basedir/view/`, though I'm not exactly sure why this is (was?).

jsilvermist: Never mind, I was thinking of using with a `base` tag vs no `base` tag, removing now...

abdonrd:
We also need to update the #2.0-preview branch.

jsilvermist:
@abdonrd See #1019 for `2.0-preview` branch updates.

jsilvermist:
Changed paths in `sw-precache-config.js` to be relative, as per https://github.com/PolymerElements/polymer-starter-kit/pull/1019#issuecomment-302828787

index.html
graynorton: @FredKSchott, @jsilvermist:

To enable the multi-build, differential serving use case (which has a lot in common with the serve-from-a-subdirectory use case), @aomarks recently added the new `basePath` build option. This option injects a `<base>` tag at build time (or updates the `href` of an existing `<base>` tag).

My suggestion is that PSK not recommend manually changing the base href in the source `index.html` file, but instead recommend using the `basePath` option to set it at build time. That would allow `polymer serve` to serve from source as it currently does without any modifications.

graynorton:
@jsilvermist, in the initial comment on this PR, you mention PSK 3.0. Are there plans for such a thing? If so, can you point me to them?

jsilvermist:
@graynorton Ah PSK naming is kinda confusing, technically the current PSK for polymer 1 is 2.0, polymer 2 is 3.0, and 1.0 was the one based on a massive gulp file. I never know when it's better to refer to it as master/2.0-preview, or 2.0/3.0, sorry for any confusion.

graynorton:
Gotcha -- thanks for clarifying, @jsilvermist.

index.html
```diff
@@ -18,10 +18,19 @@
...
     <meta name="description" content="My App description">
 
-    <link rel="icon" href="/images/favicon.ico">
+    <!--
+      If deploying to a non-root path, replace href="/" with an
+      absolute reference to the root path your application will deploy to.
+      For example: href="/polymer-starter-kit/relative-path-example/"
+
+      Note: When serving locally with `polymer serve` this must be set to href="/".
+    -->
```
graynorton: As I noted in an earlier comment on an outdated diff, I suggest we recommend that people not manually update the `href` here, but instead use the new `basePath` build option to do it at build time. (Currently it looks like that option can only be set in `polymer.json`, but I think it should be invokable from the CLI as well, via `polymer build --base-path='/polymer-starter-kit/relative-path-example/'` -- I've asked @aomarks about this.)

If we do that, then we shouldn't have to worry about the `polymer serve` case.

src/my-app.html
```diff
@@ -70,9 +70,9 @@
       <app-drawer id="drawer">
         <app-toolbar>Menu</app-toolbar>
         <iron-selector selected="[[page]]" attr-for-selected="name" class="drawer-list" role="navigation">
-          <a name="view1" href="/view1">View One</a>
-          <a name="view2" href="/view2">View Two</a>
-          <a name="view3" href="/view3">View Three</a>
+          <a name="view1" href="[[rootPath]]view1">View One</a>
+          <a name="view2" href="[[rootPath]]view2">View Two</a>
+          <a name="view3" href="[[rootPath]]view3">View Three</a>
```
graynorton: @jsilvermist, can you update the `2.0-preview` PR to include `[[rootPath]]` in the URLs there as well? As you have noted previously, it's not required for the serve-from-a-subdirectory use case in Polymer 2.0, but it is required for the multiple-build / differential serving use case.

index.html
```diff
@@ -89,12 +99,12 @@
       // Load pre-caching Service Worker
       if ('serviceWorker' in navigator) {
         window.addEventListener('load', function() {
-          navigator.serviceWorker.register('/service-worker.js');
+          navigator.serviceWorker.register('service-worker.js');
```
graynorton: I'd like to add a second argument to this `register()` call to explicitly specify the service worker's scope:
```
navigator.serviceWorker.register('service-worker.js', {scope: Polymer.rootPath});
```
This is another change that isn't necessary for the serve-from-a-subdirectory use case but allows the same template to be used for the multiple-build / differential serving use case.

We'd want the same change made to the `2.0-preview` PR.

jsilvermist: At the time of calling register for the service worker Polymer hasn't been loaded yet, so unless the user explicitly sets `Polymer.rootPath` won't this be undefined? Is it safe to move the service worker registration to after the import of `my-app` or will this cause requests to not be properly handled?

jsilvermist: Never mind, this is why I shouldn't code at 2am, just noticed the load event...

index.html
jsilvermist: Should I add a comment in place of the base tag, or will that just be documented elsewhere?

src/my-app.html
jsilvermist: Done on 2.0-preview.

index.html
jsilvermist: Done on both master and 2.0-preview.

index.html
graynorton: I think it still makes sense to have a comment here, advising people to use `--base-path` on the command line or `basePath` in `polymer.json`.

I'd lean toward leaving the base tag here as well, with `href="/"`. It doesn't hurt anything and serves to demonstrate / explain what's going on.

jsilvermist:
```html
<!--
  When deploying to a non-root path, transform the base href by adding an 
  absolute reference to the root path your application will be deployed to, by either:

  - Adding a `basePath` property to your build configuration in `polymer.json`.
    For example: { "preset": "es5-bundled", "basePath": "/subfolder/app-root/" }

  - Using the `--base-path` command line flag for polymer build in the CLI.
    For example: polymer build --base-path /subfolder/app-root/
-->
<base href="/">
```
How's this sound? I wanted to make it obvious that the `basePath` in the `polymer.json` is added inside the build config, not in the main object, but not sure if it's better to leave it as above or simplify to:
```
  - Adding a `basePath` property to your build configuration in `polymer.json`.
    For example: "basePath": "/subfolder/app-root/"
```

graynorton: I like your longer, more explicit version. Thanks!

jsilvermist: Updated comment on both `master`/`2.0-preview`.

graynorton:
@jsilvermist, there's one more change that we need to make to both the Polymer 1.x and Polymer 2.0 versions to allow multi-build / differential serving deployment to work out of the box, which is a high priority for us.

Specifically, in the differential serving case, we need to set `Polymer.rootPath` to `window.location.origin` prior to loading Polymer. Doing so of course breaks the `rootPath` behavior for the serve-from-a-subdirectory case, so we'd need to include a comment telling people to remove or comment out the line in that case.

Once we get that change in, we can update the templates in the CLI.

Does this make sense? Do you want to take a crack at this, or would you prefer that we merge the current PRs and create new PRs for this last change?

cc @aomarks @FredKSchott @abdonrd

jsilvermist:
@graynorton What exactly is the use case for the `window.location.origin` change? I'm not entirely clear on how differential serving will be implemented, assuming it's always served from the root directory, would something like this work? But then if it does work, why is this change required?
```js
<var|const> baseUrl = document.querySelector('base').href;

// If serving from root directory
if (baseUrl.pathname === '/') {
  Polymer.rootPath = window.location.origin;
}
```

graynorton:
Good question, @jsilvermist.

Unfortunately, that doesn't work.

In a differential serving deployment, multiple builds are uploaded to the server, with the resources for each build living under a named folder (e.g. 'es6-unbundled' or 'es5-bundled').

The server exposes a single, dynamic entrypoint that selects a specific build based on the requesting user-agent and performs an on-the-fly "rewrite", serving the `index.html` for the selected build (which has the base href set to the build folder, so that the relative paths to subresources resolve correctly, even though the `index.html` file is being served from a URL outside the build folder).

The difference between this and the serve-from-a-subdirectory case is that we need the app's navigation links to be specified relative to the server root, even though the static resources are served out of a subdirectory. By default, Polymer sets `rootPath` to `document.baseURI`, which is affected by the `href` of the `<base>` tag, when present. This is exactly what the serve-from-a-subdirectory case needs, but not what differential serving needs, so we need to manually override in that case.

I don't think there's anything we can do in the template to make it just "do the right thing" at runtime, so we need to either make a static tweak to the template (as I requested) or have the tools intervene. We don't favor tools intervention since this tweak is pretty specific to PSK and the desire to support serving from a subdirectory. For cases where that is not a requirement, our recommendation would be to simply specify navigation links with leading slashes, making them always relative to the server root.

Let me know if you have further questions. Thanks!

graynorton:
Hmm...now that I think of it, maybe it is possible to do something smart at runtime.

I think possibly something like this could work:

```
// If baseURI is not the same as location.href, assume we're
// doing differential serving and set rootPath accordingly
if (document.baseURI !== window.location.href ) {
  Polymer.rootPath = window.location.origin;
}
```

jsilvermist:
@graynorton In my test this check breaks due to:
On the page`https://silverlinkz.net/_test/polymer/psk3-differential/es6-unbundled/view1`

`document.baseURI` == `https://silverlinkz.net/_test/polymer/psk3-differential/es6-unbundled/`
`window.location.href` == `https://silverlinkz.net/_test/polymer/psk3-differential/es6-unbundled/view1`

Don't let the word differential in the url confuse you, I just named it that to differentiate it from my other tests, no differential serving is going on here.

---

What about using a flag to signal intent for differential serving? Something the CLI could inject?

graynorton:
Yeah, what you're seeing makes sense, so I'm back to my initial instinct that there's no easy runtime switch.

**tl;dr:** I think the simplest thing for now is just to include the manual override for `Polymer.rootPath` in the template, with a comment advising to remove it if you intend to serve from a subdirectory.

This doesn't seem too bad to me, given that the original solution (prior to adding `basePath` support to the CLI) also involved manually editing `index.html` to enable subdirectory serving.

Since the need to override `Polymer.rootPath` is quite specific to the PSK's convention for enabling subdirectory serving, I don't think there's a general-purpose CLI solution that feels good (i.e., I don't think we want to hard-code anything into `polymer-build`).

A longer-term solution, if we're not satisfied with the need to edit `index.html`, would be to maintain a separate branch for the subdirectory-serving-enabled version of the PSK; the main branch would use leading slashes for nav links and wouldn't need the `rootPattern` logic for `<app-route>`, while the subdirectory-serving-ready branch would include the `rootPattern` logic and use relative nav links without leading slashes (or `[[rootPath]]`-prefixed links, in the case of PSK v2 / Polymer 1.x).

If we went in this direction, I think we'd use the main branch for the built-in PSK generator for `polymer-init`, while making the subdirectory-serving version available as an add-on generator.

Another possibility would be to replace the built-in generator for the PSK – currently a generic one that simply fetches from GitHub – with a custom generator that prompted the user to enable or disable subdirectory serving and then switched out or automatically modified the template accordingly...but it's not clear to me that it would be worth the effort.

jsilvermist:
@graynorton Yeah while those solutions could possibly be something to look into in the future, they all sadly seem rather maintenance heavy for minimum benefit.

Is differential serving intended to also be used with Polymer 1 PSK? If so, we should probably add build presets to the `polymer.json` similar to what `2.0-preview` has. (Though currently I believe PSK for Polymer 1 is all ES5?)

I'll add this with a comment asap on Polymer 2 PSK, and on here also if wanted.

jsilvermist:
Not sure how to phrase this in the comments, it's easy if relative paths are the default:
https://github.com/jsilvermist/polymer-starter-kit/commit/0a63cf5c6ce1e47866bdc2b5d9d335bb9c1a433e
but when making it so that the default if for differential serving, I'm not sure how to keep it simple while making obvious which line they should comment out. So far I have this:
https://github.com/jsilvermist/polymer-starter-kit/commit/06225d94de58a8ad8b476d1b15a0def5b18233fc

Also, since differential builds are also technically being deployed to a non-root path with a common entrypoint, I'm not sure how to eliminate possible confusion here.

graynorton:
@jsilvermist, thanks for taking the time to work on these. I have gotten pulled into other things the last two days and am out of time for today, but I'll take a look when back on Tuesday and we can get this put to bed!

Since we've already come this far, my inclination is to go ahead and make the changes to both the Polymer 1 PSK and Polymer 2 PSK; even though Polymer 1.x is all ES5, you might still adopt differential serving to selectively serve a bundled build and an unbundled build, for example. That said, we will imminently be dropping the Polymer 1.x templates from the default list in the CLI, so I don't feel super strongly about it; the priority is definitely on the Polymer 2.x-compatible version.

graynorton:
@jsilvermist, sorry for the delay in getting back to this.

Here is my whack at some comment text; it assumes that we'll also add a bit more detail to the PSK README (which I haven't tried drafting yet). Take a look and let me know what you think...

```html
<!--
    The `<base> tag below is present to support two advanced deployment options:
    a) differential serving, and b) serving from a non-root path.

    In either case, you shouldn't generally edit this `<base>` tag yourself;
    instead, add a `basePath` property to your build configuration in `polymer.json`
    or use the `--base-path` command-line option for `polymer build`.

    For more details on differential serving and serving from a non-root path, see
    the Polymer Starter Kit README.

    Note: If you intend to serve from a non-root path, you will also need to manually
    comment out one line of JavaScript -- see [non-root-path] below.
-->
```

```html
<script>
    // [non-root-path] By default, we set `Polymer.rootPath` to the server root path
    // to support differential serving; this setting is also compatible with standard,
    // single-build deployments. If you intend to serve your app from a non-root path,
    // comment out the line below.
    window.Polymer = { rootPath: window.location.origin };

    // Load and register pre-caching Service Worker
    ...
```

jsilvermist:
@graynorton LGTM! I really like how much clearer it is, also, lmk what you think of these (minor) changes:

```html
<!--  
  The `<base>` tag below is present to support two advanced deployment options:
  1) Differential serving; 2) Serving from a non-root path;

  Instead of manually editing the `<base>` tag yourself, you should generally either:
  a) Add a `basePath` property to the build configuration in your `polymer.json`.
  b) Use the `--base-path` command-line option for `polymer build`.

  For more details on differential serving and serving from a non-root path,
  see `Differential serving` and `Deploying to a non-root path` in the README.

  Note: If you intend to serve from a non-root path, you will also need to manually
  comment out one line of JavaScript, see [polymer-root-path] below.
-->
<base href="/">
```

```js
// [polymer-root-path] By default, we set `Polymer.rootPath` to the server root path
// to support differential serving; this setting is also compatible with standard,
// single-build deployments. If you intend to serve your app from a non-root path,
// comment out the line below.
window.Polymer = { rootPath: window.location.origin };
```

```markdown
### Differential serving

-- I have no idea what to put here --

### Deploying to a non-root path

-- Or here --
```

graynorton:
@jsilvermist, your changes look good!

For the README sections, I just imagined a brief conceptual description of each type of deployment, along with whatever practical advice / steps might be helpful to someone looking to deploy in one of these ways.

I can write some text for differential serving, and would be happy to draft some for non-root-path as well, though you should absolutely go for it if you feel inspired.

Happy to proceed however you prefer. We can keep collaborating via GH comments, or if you want we can merge this PR (with README mentions removed or with stubs) and I'll open a separate PR to add the README stuff. Let me know.

abdonrd:
I think we can merge the PR and improve the docs in another PR. :)

jsilvermist:
@graynorton @abdonrd Okay sure I'll push the latest changes without the README line, and we can go ahead and merge this then. Also, will gladly help as much as possible with a new PR for the README updates!

jsilvermist:
Oh wait before merging this there's the issue from `2.0-preview` (#1020), and my PR (#1021) which adds a compatibility check and regex for `basePattern` since `new URL(...).pathname` isn't supported on IE11, how do we want to handle this? (Both here and for `2.0-preview`)

graynorton:
Your fix in #1021 looks reasonable to me. I think you could make a case for using the regex everywhere, in the interest of minimizing code size and complexity, but it is nice to use the modern feature where it's available. I think it's 👍 either way, but not having been involved in PSK style and conventions historically, I defer to others: @abdonrd, @FredKSchott, @robdodson, thoughts?

abdonrd:
@jsilvermist @graynorton I just commented in https://github.com/PolymerElements/polymer-starter-kit/pull/1021#pullrequestreview-41378137

arthurevans:
Hey folks,

After testing this, I think we have to make a couple of changes to make everything play nice together. I put together a proof-of-concept over here that serves two PSK-based apps from separate URLs using prpl-server-node, so it's doing both non-root URLs and differential serving at the same time.

https://github.com/PolymerLabs/multitenant-prpl

README lists the changes I made to the PSK installs.

Cc: @jsilvermist @graynorton

graynorton:
@arthurevans, thanks for looking at this.

My personal take is that the whole enchilada (support for both differential serving and non-root-path serving at once) is enough of a corner case that we don't want to make adopting either of the more basic use cases any more complicated.

In general, I wouldn't advocate making any changes to the template itself to support this use case, but I could see documenting the necessary steps for those who might be interested.

Though I haven't tried it out, one potential issue I see with the steps you outlined in your README is that `rootPattern` is scoped locally, so while it could be used as a prefix for the navigation links in the main side nav (as you've shown), it would have to be either bumped up into some shared scope or respecified locally to prefix any nav links that are defined outside of the main app element. (For example, see the link in the 404 view.)

arthurevans:
@graynorton Fair point. Although I think if we do things right for non-root-path, doing both is pretty easy. I _think_ the only thing that you need to do to cover this particular corner case is hard code the basePath in polymer.json ... I think all the other changes are required for non-root-path.

I may be missing something about the current handling of `rootPath`/`rootPattern` in the template. If I'm following this correctly, we're setting rootPath to (say) `https://example.com/` and then taking the pathname portion of that—`/`—as our `rootPattern`. It's not clear to me why we need to set this `rootPattern` at all; can't we just set `rootPath` to `/` and use that everywhere?

The way I did it, `rootPattern` was scoped to the app's root path `/prpl/`, so the link in the 404 view goes the right place—in this case, `/prpl/`. 

As for the other changes:

- specifying the url-space-regex is required for the non-root-path use case. If fixes the case where you have a link outside of the app (i.e., `/prpl/` links back to `/`). Without this, links outside the app fail.

- The `rootPath` => `rootPattern` change in the anchors... Probably due to my confusion above about rootPath/rootPattern. Currently, they don't work for the non-root-path case, unless you override rootPath instead of rootPattern.  (If you override rootPath to, say, `/prpl/`, you don't need this change—and I don't think you need `rootPattern` at all).

graynorton:
@arthurevans, I think you're on to something. I'm playing around with this right now, but I think it may be possible to get rid of `rootPattern` altogether, which would simplify things a bit. Will report back...

PR #1012, opened 2017-04-26
Update lint setup

abdonrd:
Now will be easy add more linters like `polymer lint` or `stylelint`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,21 @@ language: node_js
 sudo: required
 dist: trusty
 addons:
-  firefox: latest
+  firefox: "latest"
   apt:
     sources:
       - google-chrome
     packages:
       - google-chrome-stable
 node_js:
-  - node
-  - '6'
-before_script:
-  - npm install -g polymer-cli@next
+  - "node"
+  - "6"
+install:
+  - npm install
+  - npm install -g polymer-cli
   - polymer install
+before_script:
+  - npm run lint
 script:
-  - xvfb-run npm test
+  - xvfb-run npm run test
   - xvfb-run npm run test:integration
--- a/package.json
+++ b/package.json
@@ -7,8 +7,9 @@
     "eslint-plugin-html": "^2.0.0"
   },
   "scripts": {
-    "lint": "eslint . --ext js,html --ignore-path .gitignore",
-    "test": "npm run lint && polymer test",
+    "lint": "npm run lint:javascript",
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -2,18 +2,21 @@ language: node_js
...
     sources:
       - google-chrome
     packages:
       - google-chrome-stable
 node_js:
-  - node
-  - '6'
-before_script:
-  - npm install -g polymer-cli@next
+  - "node"
```
stramel: I noticed a lot of Google repos are now using `stable` over `node` here.

.travis.yml
```diff
@@ -2,18 +2,21 @@ language: node_js
...
+  - "6"
+install:
+  - npm install
+  - npm install -g polymer-cli
   - polymer install
+before_script:
+  - npm run lint
 script:
-  - xvfb-run npm test
+  - xvfb-run npm run test
```
stramel: Are you adding `run` here for consistency?

package.json
```diff
@@ -7,8 +7,9 @@
     "eslint-plugin-html": "^1.7.0"
   },
   "scripts": {
-    "lint": "eslint . --ext js,html --ignore-path .gitignore",
-    "test": "npm run lint && polymer test",
+    "lint": "npm run lint:javascript",
```
stramel: so would we can add `polymer lint` here maybe after they allow for no error on warnings?

.travis.yml
abdonrd: I'm based on the Polymer CLI: [Polymer/polymer-cli: `.travis.yml#L11-L13`](https://github.com/Polymer/polymer-cli/blob/master/.travis.yml#L11-L13)

@FredKSchott any thoughts?

.travis.yml
abdonrd: Yes, just for that. :)

package.json
abdonrd: Exactly! Something like:

```json
[...]
"lint": "npm run lint:javascript && polymer lint",
"lint:javascript": "eslint . --ext js,html --ignore-path .gitignore",
[...]
```

.travis.yml
abdonrd: In the Travis docs only `node` appears:
[Travis docs | Specifying Node.js versions](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions)

stramel: Works for me. Just had noticed a lot of them using stable. I couldn't find docs on it either.

PR #1014, opened 2017-05-15
Fix lint warnings

justinfagnani:


src/my-app.html
```diff
@@ -22,6 +22,11 @@
 <link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="my-icons.html">
 
+<link rel="lazy-import" href="my-view1.html">
```
abdonrd: In other projects I see the `lazy-import` inside the `dom-module`.
@justinfagnani What is the best practice?

jsilvermist: @justinfagnani Is this just to allow the analyzer/linter to resolve the fragments? Isn't that the point of fragments in `polymer.json`?

I feel like this should be commented since it's not a proper usage example for lazy-imports (?) (being outside of dom-module), and could easily give new users a lot of confusion.

PR #1018, opened 2017-05-16
[2.0] New: Add polymer lint to lint script

stramel:
Seems like all lint errors were fixed by @justinfagnani. It would be a shame not to go ahead and lint

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "eslint-plugin-html": "^2.0.0"
   },
   "scripts": {
-    "lint": "npm run lint:javascript",
+    "lint": "npm run lint:javascript && polymer lint",
     "lint:javascript": "eslint . --ext js,html --ignore-path .gitignore",
     "test": "polymer test",
     "test:integration": "polymer build # test that psk builds without error with the CLI"


package.json
```diff
@@ -7,7 +7,7 @@
     "eslint-plugin-html": "^2.0.0"
   },
   "scripts": {
-    "lint": "npm run lint:javascript",
+    "lint": "npm run lint:javascript && polymer lint src/*.html",
```
abdonrd: Normally we only use `polymer lint`.

stramel: @abdonrd So, `polymer lint` will lint all the files including the build directory.

Steps to Reproduce:

1. `npm i -g polymer-cli`
1. `polymer build`
1. `polymer lint`

abdonrd: Can we do something like ESLint ignore path?
`eslint . --ignore-path .gitignore`

cc @rictic

stramel: Currently, this isn't implemented, although probably would be a great feature!

stramel: @abdonrd Is this holding this PR from moving forward?

abdonrd: While waiting for https://github.com/Polymer/polymer-linter/issues/88...

As Travis do the lint first, remove the to `src/*.html` and I'll merge it!

stramel:
@abdonrd Updated!

abdonrd:
Thanks for the contribution @stramel! 👏

PR #1019, opened 2017-05-17
[2.0-preview] Bring 2.0-preview in line with polymer 1 tentative changes

jsilvermist:
Remove `baseUrl` and path concatenation to bring the 2.0-preview branch in line with the tentative polymer 1 starter kit relative path changes. REF: #1008 

**Question 1**: Should I also change `basePattern` to use a computed property?

I don't mind either constructor or computed property personally, but since it's a computed property in the polymer 1 starter kit and we already have the property defined for `polymer lint`, I'm thinking maybe we should switch it?

**~~Question 2~~**: In the live previews (listed below) I noticed the generated service worker is failing for now on relative paths, I believe this is fixed soon (?), but for now, should we change paths in `sw-precache-config.js` to relative, or are they fine since they would be rewritten the for base path anyways?

**Regarding Question 2**: Updated paths to relative in `sw-precache-config.js`.

#### Live Previews:
[es5-bundled preset](https://silverlinkz.net/_test/polymer/psk3-relative/es5-bundled/)
[es6-unbundled preset](https://silverlinkz.net/_test/polymer/psk3-relative/es6-unbundled/)

### DIFF
--- a/index.html
+++ b/index.html
@@ -65,7 +65,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // server root path to support differential serving; this setting is also
       // compatible with standard, single-build deployments. If you intend to
       // serve your app from a non-root path, comment out the line below.
-      window.Polymer.rootPath = window.location.origin;
+      window.Polymer = {rootPath: window.location.origin};
 
       // Load and register pre-caching Service Worker
       if ('serviceWorker' in navigator) {
--- a/index.html
+++ b/index.html
@@ -61,11 +61,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <meta name="msapplication-tap-highlight" content="no">
 
     <script>
+      // Use a regex for `window.location.origin` for compatibility with IE11.
+      const locationOrigin = window.location.href.match(/^.+?\/\/.*?\//)[0];
+
       // [polymer-root-path] By default, we set `Polymer.rootPath` to the
       // server root path to support differential serving; this setting is also
       // compatible with standard, single-build deployments. If you intend to
       // serve your app from a non-root path, comment out the line below.
-      window.Polymer = {rootPath: window.location.origin};
+      window.Polymer = {rootPath: locationOrigin};
 
       // Load and register pre-caching Service Worker
       if ('serviceWorker' in navigator) {


graynorton:
Regarding *Question 2*, to make the service worker work for apps not served from the server root, we will need to use relative paths -- both for the static files specified in PSK's `sw-precache-config.js`, and for the files added by polymer-build. We've been discussing this today and think this is the right direction to take. It helps in both the serve-from-a-subdirectory use case and in the multi-build, differential-serving use case.

cc @aomarks @usergenic

jsilvermist:
@graynorton Updated the paths to relative in `sw-precache-config.js`, but in the generated sw they are still transformed to absolute without any prefix prepended in `precacheConfig`.

graynorton:
@jsilvermist, yeah, we'll need to make a change on the Tools side to get all the way there. Just created this issue to track: https://github.com/Polymer/polymer-cli/issues/761

jsilvermist:
Updated base tag comment as per https://github.com/PolymerElements/polymer-starter-kit/pull/1008#discussion_r117648193.

Since this comment references a CLI `--base-path` option which doesn't exist yet, it should be added before merging this.

If you want to merge this first or not add the CLI option, I can edit the comment for now to only include the `polymer.json` method.

aomarks:
FYI here's the bug to add the --base-path flag https://github.com/Polymer/polymer-cli/issues/745

abdonrd:
@jsilvermist can you do a rebase? 🙂

abdonrd:
I merged `master` (with the changes) to `2.0-preview` and I see:

screen shot 2017-06-01 at 7 37 25 pm

jsilvermist:
@abdonrd Rebased sorry for the delay, only thing left is the change of the line you pointed to haha!

abdonrd:
Right! Haha
Thanks for all @jsilvermist!

abdonrd:
@jsilvermist @graynorton oops! I just see the `rootPath` rewrite breaks the links:

```javascript
window.Polymer = {rootPath: window.location.origin};
```

screen shot 2017-06-02 at 9 51 08 am

We also have this issue in `master` with the:

```javascript
window.Polymer.rootPath = window.location.origin;
```

abdonrd:
The solution will be add the `/`?

screen shot 2017-06-02 at 9 55 25 am

web-padawan:
FWIW, `window.location.origin` can be `undefined` in IE11 in [some cases](https://stackoverflow.com/a/38930269/6694206).

jsilvermist:
@abdonrd @graynorton @web-padawan Oh wow not sure how I didn't catch this. We should add the `/` to `Polymer.rootPath` instead of the links to maintain support for relative paths when commenting out that line.

I tested this on `http://localhost:8000` in IE11 and it seemed like the origin was set properly?

To also support IE11 (*Gah!*) we could do something like this:
```
// Use a regex for `window.location.origin` for compatibility with IE11.
const locationOrigin = window.location.href.match(/^.+?\/\/.*?\//)[0];

// [polymer-root-path] By default, we set `Polymer.rootPath` to the
// server root path to support differential serving; this setting is also
// compatible with standard, single-build deployments. If you intend to
// serve your app from a non-root path, comment out the line below.
window.Polymer = {rootPath: locationOrigin};
```
It seems to me that `window.location.href` always ends in a slash, but if this isn't the case we could use:
```
// Use a regex for `window.location.origin` for compatibility with IE11.
const pathOrigin = window.location.href.match(/^.+?\/\/[^\/$]*/)[0] + '/';

// [polymer-root-path] By default, we set `Polymer.rootPath` to the
// server root path to support differential serving; this setting is also
// compatible with standard, single-build deployments. If you intend to
// serve your app from a non-root path, comment out the line below.
window.Polymer = {rootPath: pathOrigin};
```
(Renamed location to path because of lint line length)

graynorton:
@jsilvermist, @abdonrd, I have been digging into @arthurevans's notes on the related, already-merged, Polymer 1.x-compatible PR: #1008.

I think there is a somewhat simpler solution, based on Arthur's insights, that will satisfy all of the use cases we've discussed (standard single-build deployment from server root, single-build deployment from non-root directory, multi-build differential serving from root, multi-build differential serving from non-root).

If no one objects, in the interest of getting this closed out, I'll create a separate PR tonight or first thing tomorrow morning and ask you all to review. In the meantime, let's hold off on merging this one.

cc @aomarks

jsilvermist:
@graynorton Sure! If we can get this working without `basePattern`, and without having to use all these workarounds then SGTM! I just hope we don't have to create too many static change points.

abdonrd:
@graynorton sure! Waiting! 😬

graynorton:
OK, #1030 is open.

@jsilvermist, I think it meets those criteria nicely. Same number of manual edit points as before -- just one. Instead of commenting out the `rootPath` line, you just change it to the subdirectory you intend to serve from. As for `polymer serve`, works fine if you do `polymer serve build` and then access your app at (for example) `http://localhost:8080/my-app`.

Definitely give it a look and chime in if you encounter any issues!

abdonrd:
Close in favor to #1030.

PR #1021, opened 2017-05-27
Add regex for use with IE11 instead of URL()

### LINKED ISSUE #1020: changes in my-app.html breaks polymer-starter-kit (2.0-preview) in IE11
I've identified that after the addition (0b67b51367f3754b369d00141dc604f2752ce7ee) of this part of code in src/my-app.html, polymer-starter-kit (2.0-preview) failed to work in IE11 when run with polymer serve:
```
   constructor() {
        super();

        // Get root pattern for app-route, for more info about `rootPath` see:
        // https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates
        this.rootPattern = (new URL(this.rootPath)).pathname;
      }
```
To be specific, this line caused the problem and when commented out, it works:
`        this.rootPattern = (new URL(this.rootPath)).pathname;`

jsilvermist:
So it turns out that `new URL(...)` isn't supported on IE11, this fix uses the same method on supported browsers, but uses a regex on unsupported browsers.

I wasn't sure if I should just change it to the regex, so for now I kept both with a check for compatibility and moved it all to a computed function to clean it up a bit. This also brings it in line with the pattern used on the tentative Polymer 1 PSK changes. (Will update #master PR with fix after this is finalized.)

If there's a cleaner way to do this, please lmk!

Fixes #1020

### DIFF
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -142,7 +142,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates
 
         // If URL API isn't supported (IE11), use a regex for the URL pathname.
-        if (!URL.prototype || !('pathname' in URL.prototype)) {
+        if (!(URL.prototype && 'pathname' in URL.prototype)) {
           return rootPath.replace(/^.+?\/\/?.*?\/|$/, '/');
         }
 
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -140,8 +140,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       computeRootPattern(rootPath) {
         // Get root pattern for app-route, for more info about `rootPath` see:
         // https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates
-        if (window.URL && window.URL.prototype && ('href' in window.URL.prototype)) {
-          // Use `new URL()` if supported by the browser.
+        if (window.URL && window.URL.prototype &&
+            ('href' in window.URL.prototype)) {
           return (new URL(rootPath)).pathname;
         } else {
           // Use a regex if `new URL()` isn't supported. (IE11)
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -140,13 +140,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       computeRootPattern(rootPath) {
         // Get root pattern for app-route, for more info about `rootPath` see:
         // https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates
-        if (window.URL && window.URL.prototype &&
-            ('href' in window.URL.prototype)) {
-          return (new URL(rootPath)).pathname;
-        } else {
-          // Use a regex if `new URL()` isn't supported. (IE11)
-          return rootPath.replace(/^(.+?\/\/)?(.*?)(\/|$)/, '/');
+
+        // Use a regex if URL API isn't supported. (IE11)
+        if (!URL.prototype || !('pathname' in URL.prototype)) {
+          return rootPath.replace(/^.+?\/\/?.*?\/|$/, '/');
         }
+
+        return (new URL(rootPath)).pathname;
       }
 
       _routePageChanged(page) {
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -122,7 +122,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             reflectToAttribute: true,
             observer: '_pageChanged',
           },
-          rootPattern: String,
+          rootPattern: {
+            type: String,
+            computed: 'computeRootPattern(rootPath)',
+          },
           routeData: Object,
           subroute: String,
         };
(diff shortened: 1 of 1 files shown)


abdonrd:
Hummm... I tried the URL API in IE11 and it works:

screen shot 2017-06-01 at 12 34 02 am

src/my-app.html
```diff
@@ -134,12 +137,16 @@
...
       }
 
-      constructor() {
-        super();
-
+      computeRootPattern(rootPath) {
         // Get root pattern for app-route, for more info about `rootPath` see:
         // https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates
-        this.rootPattern = (new URL(this.rootPath)).pathname;
+        if (window.URL && window.URL.prototype &&
```
abdonrd: Would not be this enough?

```javascript
if (!('URL' in window)) {}
```

src/my-app.html
```diff
@@ -134,12 +137,16 @@
...
+            ('href' in window.URL.prototype)) {
+          return (new URL(rootPath)).pathname;
+        } else {
+          // Use a regex if `new URL()` isn't supported. (IE11)
+          return rootPath.replace(/^(.+?\/\/)?(.*?)(\/|$)/, '/');
```
abdonrd: We can avoid the groups to do less steps:

```javascript
return rootPath.replace(/^.+?\/\/?.*?\/|$/, '/');
```

abdonrd: And maybe remove the else:

```javascript
// Use a regex if URL API isn't supported. (IE11)
if (!('URL' in window)) {
  return rootPath.replace(/^.+?\/\/?.*?\/|$/, '/');
}

return (new URL(rootPath)).pathname;
```

src/my-app.html
jsilvermist: This check is saying it's supported, but I'm still getting failures on the URL.

ie11-nonewurl

jsilvermist: Though, I guess since `URL` is also in IE11, this could be simplified to:
```
if (!window.URL.prototype || !('pathname' in window.URL.prototype)) {...}
```
or the even simpler (but possibly less safe?):
```
if (!window.URL.prototype) {...}
```

src/my-app.html
jsilvermist: Sure, cleaned up the pattern and regex, the reason I put it in groups was that I wanted it to be as humanly readable as possible.

jsilvermist:
Not sure exactly how it's working for you, I'm still getting `Object doesn't support this action`.
Also, according to MDN this isn't supported in IE11:
https://developer.mozilla.org/en/docs/Web/API/URL#Browser_compatibility

ie11-nonewurl

abdonrd:
It's magic haha 

Now I use the Windows 8.1 image with IE11:

screen shot 2017-06-01 at 12 10 37 pm

src/my-app.html
```diff
@@ -134,12 +137,16 @@
...
-      constructor() {
-        super();
-
+      computeRootPattern(rootPath) {
         // Get root pattern for app-route, for more info about `rootPath` see:
         // https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates
-        this.rootPattern = (new URL(this.rootPath)).pathname;
+
+        // Use a regex if URL API isn't supported. (IE11)
+        if (!URL.prototype || !('pathname' in URL.prototype)) {
```
abdonrd: Inspiring in `webcomponents-loader.js`... It works in your IE11?

```javascript
if (!('pathname' in URL.prototype)) {}
```

src/my-app.html
```diff
@@ -134,12 +137,16 @@
...
 
-      constructor() {
-        super();
-
+      computeRootPattern(rootPath) {
         // Get root pattern for app-route, for more info about `rootPath` see:
         // https://www.polymer-project.org/2.0/docs/upgrade#urls-in-templates
-        this.rootPattern = (new URL(this.rootPath)).pathname;
+
+        // Use a regex if URL API isn't supported. (IE11)
```
abdonrd: I just reread it and I think we could be a bit more explicit:

`// If URL API isn't supported (IE11), use a regex to get the URL pathname.`

WDYT? 😄

src/my-app.html
jsilvermist: When I tried it before it gave:
```
Invalid operand to 'in': Object expected
```

src/my-app.html
jsilvermist: LGTM! Changing now...

jsilvermist: 2 characters too long for ESLint. Changed to:
```
// If URL API isn't supported (IE11), use a regex for the URL pathname.
```
This look ok?

src/my-app.html
abdonrd: Wow, `URL.prototype` is not an object in IE11? 😮

src/my-app.html
abdonrd: Ok! Thanks!

src/my-app.html
abdonrd: In my magic IE11 v11.0.9600.18666:

screen shot 2017-06-01 at 2 34 20 pm

jsilvermist: Ok now I'm beyond confused, that relative starter kit page you're on is using Polymer 1.9.1, if I go there then adding this to the console works too, so I guess there's some URL polyfill being added?

Though that version wasn't build using babel or anything other than just a simple `polymer build` with no presets.

Try the same code at eg [Google.com](https://google.com) and it always fails.
Also, it doesn't seem to be working with 2.0-preview regardless.

❤ Internet Explorer :sob::sob::sob:

abdonrd: Oh! The `webcomponentsjs < v1.0.0` use a URL API polyfill!
https://github.com/webcomponents/webcomponentsjs/pull/679

abdonrd: Then I think we can add this only to the Polymer v2 version:

```javascript
if (!(URL.prototype && 'pathname' in URL.prototype)) {}
```

jsilvermist: Updated compatibility check, if `master` has the `URL` polyfill from `webcomponentsjs`, then I guess #1008 should be good to go, barring the fact that it mentions using a `--base-path` command-line option for `polymer build` which I don't believe exists yet.

abdonrd: Yes, Polymer v1 uses `"webcomponentsjs": "^0.7.24"` that has the URL polyfill.

abdonrd:
Thanks @jsilvermist! 👏

PR #1027, opened 2017-06-01
Migrate PSK to Polymer 2.0

abdonrd:
This is a work in progress pull request. :)


FluorescentHallucinogen:
@abdonrd Have you tried to build it?

I've tried to build current [`2.0-preview`](https://github.com/PolymerElements/polymer-starter-kit/tree/0c12fc9901142afae027ea5e554af700f92b1d63) and [`master`](https://github.com/PolymerElements/polymer-starter-kit/tree/369af2b968c77a730c453aae5f9dc3b580e874c3) branches, both do not work.

I believe the problem is in https://github.com/PolymerElements/polymer-starter-kit/pull/1008.

abdonrd:
@FluorescentHallucinogen yes, it doesn't work right now. It is work in progress.

References to https://github.com/PolymerElements/polymer-starter-kit/pull/1008 and https://github.com/PolymerElements/polymer-starter-kit/pull/1019 about the relative paths.

src/my-app.html
```diff
@@ -101,49 +111,78 @@
...
+        };
+      }
+
+      static get observers() {
+        return [
+          '_routePageChanged(routeData.page)',
+        ];
+      }
+
+      computeRootPattern(rootPath) {
```
danizord: Should be prefixed with `_`?

src/my-app.html
```diff
@@ -22,6 +22,11 @@
 <link rel="import" href="../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="my-icons.html">
 
+<link rel="lazy-import" href="my-view1.html">
+<link rel="lazy-import" href="my-view2.html">
+<link rel="lazy-import" href="my-view3.html">
+<link rel="lazy-import" href="my-view404.html">
```
danizord: Should not it be inside template?

abdonrd: Nop! The [lazy-imports](https://github.com/Polymer/lazy-imports) has the `lazy-import` outside the `template`.

src/my-app.html
abdonrd: Probably. But first we are waiting for https://github.com/PolymerElements/polymer-starter-kit/pull/1019#issuecomment-306350626.

abdonrd:
We probably merge after #998.

PR #1030, opened 2017-06-06
Simpler support for differential serving, non-root-path deployments

graynorton:
This PR, based on @arthurevans's comments on #1008, is a simpler implementation of support for differential serving and non-root-path deployments.

The key difference is that we set `Polymer.rootPath` to a server-relative path with a leading slash, instead of mimicking its "natural" form (a full URL). We default to `/`, which works for differential serving and for standard, single-build deployments from the server root. For non-root-path deployments, devs will manually edit it to specify the path from which they'll serve (e.g. `/my-app/`).

This eliminates the need for the computed `rootPattern` property, since that property was just massaging `rootPath` into the form that we're now using directly.

Also: added `url-space-regex` property to `<app-location>` to filter out URL changes not beneath the app's root path, so that the app's router won't hijack navigations attempting to move outside the app.

Would still like to provide more detailed info in the README, but saving that for a separate PR.

This PR should supersede #1019.

cc @jsilvermist

### DIFF
--- a/index.html
+++ b/index.html
@@ -27,7 +27,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       b) Use the `--base-path` command-line option for `polymer build`.
 
       Note: If you intend to serve from a non-root path, you will also need to manually
-      comment out one line of JavaScript, see [polymer-root-path] below.
+      edit one line of JavaScript, see [polymer-root-path] below.
     -->
     <base href="/">
 
@@ -62,10 +62,15 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     <script>
       // [polymer-root-path] By default, we set `Polymer.rootPath` to the
-      // server root path to support differential serving; this setting is also
-      // compatible with standard, single-build deployments. If you intend to
-      // serve your app from a non-root path, comment out the line below.
-      window.Polymer.rootPath = window.location.origin;
+      // server root path (`/`). Leave this line unchanged if you intend to serve
+      // your app from the root path (e.g., if users will access your app from
+      // URLs like `https://my.domain/`, `https://my.domain/view1`, etc.).
+      //
+      // If you intend to serve your app from a non-root path (e.g., with URLs
+      // like `https://my.domain/my-app/` and `https://my.domain/my-app/view1`),
+      // edit this line to indicate the path from which you'll be serving (e.g.
+      // `/my-app/`).
+      window.Polymer = {rootPath: '/'};
 
       // Load and register pre-caching Service Worker
       if ('serviceWorker' in navigator) {
--- a/index.html
+++ b/index.html
@@ -26,8 +26,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       a) Add a `basePath` property to the build configuration in your `polymer.json`.
       b) Use the `--base-path` command-line option for `polymer build`.
 
-      Note: If you intend to serve from a non-root path, you will also need to manually
-      edit one line of JavaScript, see [polymer-root-path] below.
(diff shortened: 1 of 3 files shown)


arthurevans:
LGTM. 

The other thing I encountered isn't directly linked to the root path, and I think it's a documentation-only issue. Basically, if your server root != the build directory, you need to manually configure the base path for each build in `polymer.json`. You can encounter this, for example, if you use prpl-server-node as a library to serve a non-root-path Polymer app as well as other resources.

Not sure how common this pattern is, but I think it may go hand-in-hand with serving an app from a non-root path in some scenarios, so we should cover it in docs.

jsilvermist:
Looks good to me, I just wish there was a way to dynamically change `rootPath` like we do the `basePath`, but it seems any way to do that would just end up adding more complexity not less.

graynorton:
@jsilvermist, I think there's a possibility of adding tools support for setting `rootPath` at some point, but it's not immediately obvious whether it would make sense or exactly how it would work. I feel like this is a decent start.

jsilvermist:
@graynorton Agreed!

index.html
```diff
@@ -27,7 +27,7 @@
       b) Use the `--base-path` command-line option for `polymer build`.
 
       Note: If you intend to serve from a non-root path, you will also need to manually
-      comment out one line of JavaScript, see [polymer-root-path] below.
+      edit one line of JavaScript, see [polymer-root-path] below.
```
abdonrd: I propose update the last note to only:

`Note: If you intend to serve from a non-root path, see [polymer-root-path] below.`

index.html
```diff
@@ -62,10 +62,15 @@
...
-      // serve your app from a non-root path, comment out the line below.
-      window.Polymer.rootPath = window.location.origin;
+      // server root path (`/`). Leave this line unchanged if you intend to serve
+      // your app from the root path (e.g., if users will access your app from
+      // URLs like `https://my.domain/`, `https://my.domain/view1`, etc.).
+      //
+      // If you intend to serve your app from a non-root path (e.g., with URLs
+      // like `https://my.domain/my-app/` and `https://my.domain/my-app/view1`),
+      // edit this line to indicate the path from which you'll be serving (e.g.
+      // `/my-app/`).
```
abdonrd: I would like to propose a minor doc change:

```javascript
/**
 * [polymer-root-path]
 *
 * By default, we set `Polymer.rootPath` to the server root path (`/`).
 * Leave this line unchanged if you intend to serve your app from the root
 * path (e.g., with URLs like `my.domain/`, `my.domain/view1`, etc).
 *
 * If you intend to serve your app from a non-root path (e.g., with URLs
 * like `my.domain/my-app/` and `my.domain/my-app/view1`), edit this line
 * to indicate the path from which you'll be serving (e.g. `/my-app/`).
 */
```

abdonrd:
Also, when I run:

`polymer build`
`polymer serve build/es6-unbundled/`

Serving the build doesn't work by default

graynorton:
Thanks, @abdonrd -- I've made the changes you suggested. Good catch on removing `basePath: true` from the default build config.

abdonrd:
Yay! Thanks @graynorton!

abdonrd:
@graynorton do you want to send another PR to master? To get the relative paths working in the Polymer v1 version.

graynorton:
@abdonrd, I'd be happy to, but I don't think I'm likely to have time until tonight or tomorrow morning. If someone else beats me to it, I would be fine with that. :-)

cc @jsilvermist @arthurevans

jsilvermist:
@graynorton @abdonrd @arthurevans Created a Polymer 1.0 parody at #1031.

Note: Added a few comments/questions about the `url-space-regex` attribute.

PR #1031, opened 2017-06-07
[master/polymer-1] Simpler differential serving & non-root-paths

jsilvermist:
Polymer 1 parody of #1030.

### DIFF
--- a/index.html
+++ b/index.html
@@ -26,8 +26,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       a) Add a `basePath` property to the build configuration in your `polymer.json`.
       b) Use the `--base-path` command-line option for `polymer build`.
 
-      Note: If you intend to serve from a non-root path, you will also need to manually
-      comment out one line of JavaScript, see [polymer-root-path] below.
+      Note: If you intend to serve from a non-root path, see [polymer-root-path] below.
     -->
     <base href="/">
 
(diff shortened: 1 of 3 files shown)


jsilvermist:
The `url-space-regex="[[rootPath]]*"` was causing the error:
```
Uncaught SyntaxError: Invalid regular expression: /*/: Nothing to repeat
```
This error doesn't seem to be present on `2.0-preview`, but won't it still just be repeating the last `/`?

jsilvermist:
`iron-location`'s `_urlSpaceRegExp` property is equal to `/\/.*/` after this change, so it seems to be properly escaping the `/`.

On `2.0-preview` it's `/\/*/`, should I create a PR to also change this on `2.0-preview`?

graynorton:
Thanks, @jsilvermist. I have to admit that I didn't scrutinize the expression -- I lifted it straight from @arthurevans's example and did only a very cursory check to make sure that nav links weren't entirely broken. :-P

Looking at the docs for `url-space-regex`, it seems like just `/^[[rootPath]]/` (i.e., `url-space-regex="^[[rootPath]]"` ought to work.

Yes, please do file a corresponding PR against `2.0-preview` as well.

arthurevans:
Yep, my bad, I kind of just poked at it. app-location docs say:
            
This regexp is given everything after the origin in an absolute
URL. So to match just URLs that start with /search/ do:
     url-space-regex="^/search/"

Which suggests that `"^[[rootPath]]"` should work for rootPaths like `"/"` `"/foo/"` or `""`.

jsilvermist:
Updated this and sent PR for `2.0-preview` fix at #1032.

PR #1042, opened 2017-06-27
Update polymer init command

ernsheong:


abdonrd:
@justinfagnani @FredKSchott we should update the PSK README? Or the Polymer CLI should revert to `starter-kit`? Now it only have the Polymer 2 versions.

ernsheong:
Anyone?

abdonrd:
@ernsheong I'm waiting for Justin and/or Fred.

justinfagnani:
Since we pull starter kit directly from Github, let's do this update for now, then move to plain `starter-kit` with a change to CLI (and reverting this).

justinfagnani:
So LGTM

ernsheong:
Thanks. I imaging this would be the starting point for most newbies to Polymer. So it's kinda crucial to update the doc to ensure a smooth onboarding...

abdonrd:
Thanks both!

PR #1045, opened 2017-07-06
Allow swipe open the drawer

FluorescentHallucinogen:


jsilvermist:
I feel like this is a good default as it encourages a behaviour users expect from a native application, so we should probably also add this to `2.0-preview`.

abdonrd:
@jsilvermist If we merge, I will merge `master` into `2.0-preview`. :)

PR #1048, opened 2017-07-19
Remove unneeded check

abdonrd:
I think this had to be during the preview, because now it's never `undefined `.

Was added by @FredKSchott at #967.

### DIFF
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -136,12 +136,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
 
       _routePageChanged(page) {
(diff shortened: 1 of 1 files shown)


justinfagnani:
I suspect this was done because multi-prop observers can now see undefined. @FredKSchott did you actually see `undefined` here at some point?

LGTM though. you can wait for Fred for a bit if you want.

jsilvermist:
I remember seeing undefined and a white screen when testing this during the preview, it happened on load depending on the active route, but I don't remember what the exact conditions were anymore.

abdonrd:
@jsilvermist I also saw it in the preview. But it should have changed in the final release.

PR #1049, opened 2017-07-24
Add scope and use utm_source in manifest.json

FluorescentHallucinogen:


FluorescentHallucinogen:
@abdonrd PTAL.

abdonrd:
I can merge a PR with only the `start_url`, but I'm not sure about `scope`.

Anyway, I'm not sure if change to `./?utm_source=homescreen`.

FluorescentHallucinogen:
@abdonrd 
> I'm not sure about `scope`.

The scope does two important things to maintain the "app experience" we are looking for as developers.

1. Navigate outside the app: if you navigate to a URL (like click on an anchor) that is outside the scope of the app, then it will open that page in the standard browser (whatever your default browser is, even if it’s a different one).
2. Navigate into the app: this is what we call "deep linking" when it’s to a native app. But if you navigate to a URL within the scope of the manifest, that URL will be opened inside the installable app context. This means a web page, or even native app can open up your web app.

@abdonrd 
> I'm not sure if change to `./?utm_source=homescreen`.

`sw-precache` finds matching cache entries by doing a comparison with the full request URL. It's common for sites to support URL query parameters that don't affect the site's content and should be effectively ignored for the purposes of cache matching. One example is the [`utm_`-prefixed](https://support.google.com/analytics/answer/1033867) parameters used for tracking campaign performance. By default, `sw-precache` ignores `/^utm_/` parametrs.

web-padawan:
PSK is a developer boilerplate, why should it include utm parameters? 
From my point pf view, adding the stuff like this would confuse some newbies.

FluorescentHallucinogen:
@web-padawan Please take a look at `manifest.json`. At this moment PSK using non-standard `/?homescreen=1` parameter to track visits of users that added the web app to home screen. But unlike the `./?utm_source=homescreen`, `/?homescreen=1` is not ignored by `sw-precache` node module. I.e. nothing has changed for newbies.

FluorescentHallucinogen:
@abdonrd Oh, I read your message wrong. I can change `web_app_manifest` value to `homescreen`, if you want. I don't know what is better. What do you think?

abdonrd:
@FluorescentHallucinogen @web-padawan Thanks for your feedback!

I will wait another approval from the Polymer Team.

FluorescentHallucinogen:
Any progress?

abdonrd:
Waiting for the Polymer Team...

justinfagnani:
Also @FluorescentHallucinogen all PRs really should have a description that explains why the change is being made. You do that in the discussion, but it's easier if this is done up front. Thanks!

FluorescentHallucinogen:
@justinfagnani @robdodson @abdonrd FYI, Google Santa Tracker uses `utm_source=web_app_manifest` in production: https://santatracker.google.com/manifest.json.

robdodson:
I don't think you _need_ utm source. That's just if you want to do your own tracking. I'm assuming santa tracker was doing this to measure the number of PWA installs.

jsilvermist:
If `"start_url": "./?homescreen=1"` affects the service worker and there's a built in exclusion for `"start_url": "./?utm_source=web_app_manifest"` then it seems like a worthy change to me.

justinfagnani:
So can't we remove the `?homescreen=1`? Is it documented anywhere what that's for?

FluorescentHallucinogen:
@justinfagnani 
> So can't we remove the `?homescreen=1`? Is it documented anywhere what that's for?

This is for tracking visits of users that added the web app to home screen.

It is documented here: https://developers.google.com/web/updates/2015/10/display-mode.

FluorescentHallucinogen:
IMO, I believe tracking the number of PWA installs is a good practice. PSK is a developer boilerplate which combines the best practices. That's why it was added to PSK. I just suggest using standard `?utm_source=web_app_manifest` instead of non-standard `?homescreen=1` for this.

robdodson:
Reading over this thread again, I didn't notice that utm source was ignored by sw-precache. That actually is useful because you might visit the page online, add it to your homescreen, go offline, and then try to open the homescreen version. In that scenario, you would get the offline dinosaur because the service worker thinks /?homescreen=1 is different from /.

I think I'm ok with this change going in. But I think it should be documented in the PSK readme.

PR #1063, opened 2017-09-24
Use HTTPS

FluorescentHallucinogen:


jsilvermist:
Note: I believe that if these license headers are changed they will be doubled as they won't be deduplicated against the headers from all Polymer elements when building/bundling without changing those as well.

FluorescentHallucinogen:
@jsilvermist What do you mean? All comments (license headers) are stripped during build process.

jsilvermist:
I believe there's an exception for jsdoc comments beginning with `@license`, which are duplicated but not completely removed.

abdonrd:
Thanks @FluorescentHallucinogen, but @jsilvermist is right!

PR #1066, opened 2017-09-29
Cache the polyfills in runtimeCaching

abdonrd:
As @frankiefu recommended here: https://github.com/Polymer/news/pull/141#discussion_r141770890

Also, add images to `staticFileGlobs`.

### DIFF
--- a/sw-precache-config.js
+++ b/sw-precache-config.js
@@ -12,9 +12,21 @@
 
 module.exports = {
   staticFileGlobs: [
+    'bower_components/webcomponentsjs/webcomponents-loader.js',
+    'images/*',
     'index.html',
     'manifest.json',
-    'bower_components/webcomponentsjs/*',
+  ],
+  runtimeCaching: [
+    {
+      urlPattern: /bower_components\/webcomponentsjs\/.*.js/,
+      handler: 'fastest',
+      options: {
+        cache: {
+          name: 'webcomponentsjs-polyfills-cache',
+        },
+      },
+    },
   ],
   navigateFallback: 'index.html',
 };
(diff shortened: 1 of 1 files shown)


abdonrd:
About the `index.html`, should it be in `staticFileGlobs` or in `navigateFallback`? Or in both? Or in none?

- [news](https://github.com/Polymer/news): in none.
- [shop](https://github.com/Polymer/shop): in `staticFileGlobs`.
- polymer-starter-kit: in `staticFileGlobs` & `navigateFallback`.

sw-precache-config.js
```diff
@@ -12,9 +12,19 @@
...
-    'index.html',
+    'bower_components/webcomponentsjs/webcomponents-loader.js',
+    'images/*',
     'manifest.json',
-    'bower_components/webcomponentsjs/*',
   ],
-  navigateFallback: 'index.html',
+  runtimeCaching: [
+    {
+      urlPattern: /bower_components\/webcomponentsjs\/.*.js/,
```
keanulee: Let's keep the leading slash, so it doesn't match `foobower_components/`.

```js
      urlPattern: /\/bower_components\/webcomponentsjs\/.*.js/,
```

abdonrd: Maybe `^` would be better?

```javascript
urlPattern: /^bower_components\/webcomponentsjs\/.*.js/,
```

To maintain consistency, since the entire project has relative paths.

jsilvermist: @abdonrd Wouldn't that break relative paths though since it would match `/bower_components/` but not `/subdirectory/bower_components/`? (Not exactly sure what this is being matched against)

PR #1081, opened 2017-11-14
Update my-app.html

nika-begiashvili:
use let instead of var

### DIFF
--- a/src/my-app.html
+++ b/src/my-app.html
@@ -157,7 +157,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _pageChanged(page) {
         // Load page import on demand. Show 404 page if fails
-        var resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
+        let resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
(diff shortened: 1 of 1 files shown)


src/my-app.html
```diff
@@ -157,7 +157,7 @@
 
       _pageChanged(page) {
         // Load page import on demand. Show 404 page if fails
-        var resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
+        let resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
```
nidelson: If the reference (resolvedPageUrl) will not be changed, I would use it as constant.
```js
const resolvedPageUrl = this.resolveUrl('my-' + page + '.html');
```

abdonrd: Yep! 👍

nika-begiashvili: yes that's more like it

nidelson: @nika-begiashvili Update the PR for @abdonrd to make the merge for us.

abdonrd:
Thanks @nika-begiashvili & @nidelson! 👏

PR #1096, opened 2018-04-18
Updates for Polymer 3

bicknellr:
- Paths are now name specifiers.
- Removes `bower.json`.
- License updates which need to be confirmed.
  - I pulled LICENSE from one of the tools repos because this removes `bower.json`, which previously linked to the license text.
  - [x] Confirm license text.
- [x] Fix `polymer test`; possibly related to `polymer serve` starting in 'component' mode?
- [x] Confirm URL format for script tags in HTML. (`./node_modules/dep` vs `../dep`)


LICENSE
```diff
@@ -0,0 +1,29 @@
...
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
keanulee: Remove this file - canonical license file is already referenced at the top of source files.

polymer.json
```diff
@@ -2,7 +2,8 @@
   "entrypoint": "index.html",
   "shell": "src/my-app.js",
   "sources": [
-    "images/**/*"
+    "images/**/*",
+    "src/**/*"
```
keanulee: Is this necessary? I've been removing them in other 3.0 projects since it didn't seem to do anything (build discovers the tree through import statements).

LICENSE
bicknellr: Are we leaving the `license` field in `package.json`?

keanulee: Yes, that's fine (needed to suppress npm warning).

polymer.json
bicknellr: @usergenic might also know.

README.md
```diff
@@ -79,14 +79,14 @@ Run `polymer help build` for the full list of available options and optimization
 
 This command serves your app. Replace `build-folder-name` with the folder name of the build you want to serve.
 
-    polymer serve build/build-folder-name/
+    npm start build/build-folder-name/
```
notwaldorf: is this true? i.e. have you tested this

index.html
```diff
@@ -85,10 +85,10 @@
     </script>
 
     <!-- Load webcomponents-loader.js to check and load any polyfills your browser needs -->
-    <script src="./node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+    <script src="node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 
     <!-- Load your application shell -->
-    <script type="module" src="./src/my-app.js"></script>
+    <script type="module" src="src/my-app.js"></script>
```
notwaldorf: why are the relative paths wrong?

src/my-app.js
```diff
@@ -1,112 +1,114 @@
...
-@license
-Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
-This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
-The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
-The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
-Code distributed by Google as part of the polymer project is also
-subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
-*/
+ * @license
+ * Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
```
notwaldorf: blehhhh pinging @azakus to double check we should be bumping this year (i guess it's technically a new file?)

index.html
abdonrd: Should we set it like this and use the base tag to support the PRPL Server?

README.md
abdonrd: Yep!

screen shot 2018-04-19 at 14 41 09

index.html
keanulee: Don't think this change makes a difference since it's still a relative URL that uses the base tag. No action needed.

src/my-app.js
keanulee: Substantial edits have been made, so updating is fine.

src/my-app.js
```diff
@@ -1,112 +1,114 @@
...
+import '@polymer/paper-icon-button/paper-icon-button.js';
+import './my-icons.js';
 
+/**
+ * Gesture events like tap and track generated from touch will not be
+ * preventable, allowing for better scrolling performance.
+ */
 setPassiveTouchGestures(true);
+
 setRootPath(Polymer.rootPath);
```
bicknellr: `Polymer` doesn't seem to be getting imported from anywhere.

README.md
```diff
@@ -45,18 +45,18 @@ Install [Polymer CLI](https://github.com/Polymer/polymer-cli) using
...
 
-    polymer serve
+    npm start
 
 ### Build
 
-The `polymer build` command builds your Polymer application for production, using build configuration options provided by the command line or in your project's `polymer.json` file.
+The `npm run build` command builds your Polymer application for production, using build configuration options provided by the command line or in your project's `polymer.json` file.
 
 You can configure your `polymer.json` file to create multiple builds. This is necessary if you will be serving different builds optimized for different browsers. You can define your own named builds, or use presets. See the documentation on [building your project for production](https://www.polymer-project.org/2.0/toolbox/build-for-production) for more information.
```
bicknellr: I should find the new link for https://www.polymer-project.org/2.0/toolbox/build-for-production.

src/my-app.js
```diff
@@ -1,112 +1,114 @@
...
+            <my-view2 name="view2"></my-view2>
+            <my-view3 name="view3"></my-view3>
+            <my-view404 name="view404"></my-view404>
+          </iron-pages>
+        </app-header-layout>
+      </app-drawer-layout>
+    `;
   }
 
   static get is() { return 'my-app'; }
```
bicknellr: I don't think the `is` getter is necessary anymore.

README.md
bicknellr: @abdonrd added these as npm scripts in `package.json`, they use the cli there, and I think this means that you don't have to install the cli globally to run through these instructions anymore.

abdonrd: @bicknellr exactly! Use the Polymer CLI installed in the project.

src/my-app.js
bicknellr: Fixed.

src/my-app.js
bicknellr: It's set as a global in `index.html`. I'm going to skip doing anything about that in this PR but I think these might make sense to move out to `index.html` too? I've opened a branch (`global-settings`) to look at this.

src/my-app.js
azakus: Only new new files need a new date AFAIK

bicknellr: I've changed the dates back.

README.md
notwaldorf: 👍

bicknellr:
FWIW, I think googlebot is worried that my merge commit which has my personal email attached to it (thanks to GitHub not letting me choose which to use on which repos) is somebody new.

polymer.json
usergenic: If the source file is imported somewhere with a string literal identifier (i.e. not by concatenation or variable reference) then you *shouldn't* need them here.  However this is good if your source filenames can not be statically resolved by build.

bicknellr: I'm going to revert this then - it doesn't seem to be affecting anything yet. Also, I talked w/ @usergenic off of GitHub and he said that `shell` is one of the roots of this graph and dynamic imports are considered when finding other sources. Given that `src/my-app.js` eventually gets to all of the files in `src` through both regular and dynamic imports, it should be sufficient to get the whole graph. This should be easy to add back later if we find it's better.

bicknellr:
Chill out googlebot, it's still just me.

PR #1097, opened 2018-04-18
Polymer 3 project update

abdonrd:
I split the commits to easy review.


bicknellr:
Awesome! I'm going to go ahead and merge this since it's a PR to another unreleased branch and we can just continue fixing up things there. Why remove eslint though?

abdonrd:
@bicknellr  We only add it to try to make easier the review of PRs. But the dependencies were a bit outdated. And now that everything is JS, possibly something like [prettier](https://prettier.io) would make it even simpler and automatic.

PR #1098, opened 2018-04-21
3.0 preview update

abdonrd:
Some minor changes:

* Remove last ESLint presence
* Update Travis setup as Polymer/polymer
* Improve the docs


src/my-app.js
```diff
@@ -134,27 +134,36 @@ class MyApp extends PolymerElement {
   }
 
   _routePageChanged(page) {
+    /**
```
bicknellr: I don't think this should be marked as `/**` since it's not something you would expect to show up in documentation, if we had a documentation generator.

src/my-app.js
```diff
@@ -134,27 +134,36 @@ class MyApp extends PolymerElement {
...
-      // Default to 'view1' in that case.
       this.page = 'view1';
     } else if (['view1', 'view2', 'view3'].indexOf(page) !== -1) {
       this.page = page;
     } else {
       this.page = 'view404';
     }
 
-    // Close a non-persistent drawer when the page & route are changed.
+    /**
```
bicknellr: Also probably should have `/**`.

src/my-app.js
```diff
@@ -134,27 +134,36 @@ class MyApp extends PolymerElement {
...
       this.$.drawer.close();
     }
   }
 
   _pageChanged(page) {
-    // Load page import on demand. Show 404 page if fails
-    // Note: `polymer build` doesn't like string concatenation in
-    // the import statement, so break it up.
-    switch(page) {
+    /**
```
bicknellr: Also `/**`.

bicknellr:
You'll want to merge 3.0-preview into here again; the tools / cli are having some trouble with the recently released Node.js 10 so we're pinning a bunch of things to 9 for now.

abdonrd:
@bicknellr done! :)

abdonrd:
Green Travis! 🎉

.travis.yml
```diff
@@ -1,16 +1,10 @@
 language: node_js
+node_js: "9"
```
bicknellr: Could you put 6 back in here too? I think we should keep it since we're supporting it with the CLI as far as I know.

abdonrd: Oh, okey! I relied on [Polymer/polymer/.travis.yml](https://github.com/Polymer/polymer/blob/master/.travis.yml)

abdonrd: Done!

bicknellr: Thanks! I think since we're specifically expecting folks to use the tools when they're working with this repo, we should test using the same versions that we're supporting with the tools. However, for Polymer we're only really using the tools to test it. (Although maybe this is different now that we're distributing on npm?) That's sort of the way I've been thinking about it at least.

PR #1101, opened 2018-04-24
Add prettier example

abdonrd:
An example of how we can add prettier as automatic formatter.

With the `"precommit": "pretty-quick --staged",`, before each commit, `prettier` will change the format of all the staged files.

Feel free to close the PR it if it does not seem appropriate.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,3 @@
+{
+  "singleQuote": true
+}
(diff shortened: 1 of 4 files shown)


bicknellr:
I'd prefer to work on just Polymer 3 -porting and avoid adding new features at the moment but you should hold on to this PR for later - having a formatter working by default in PSK would be a nice improvement. We may as well take the changes that it made to `src/my-app.js` though. :)

PR #1106, opened 2018-04-27
Change Travis config `sudo` to `false`.

bicknellr:
I don't think this is necessary anymore: Polymer/polymer@3.x doesn't include it in its `.travis.yml` and this one is pretty much a subset of that one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ node_js:
   - "9"
   - "6"
 dist: trusty
-sudo: required
+sudo: false
 addons:
   firefox: latest
   chrome: stable


bicknellr:
I guess not. Selenium doesn't seem to be able to start Chrome, so I'm assuming that's what it was needed for? I've restarted the last build on 3.0-preview just to make sure it's not due to some current state of Travis.

bicknellr:
Yup, 3.0-preview passes and this change makes it fail.

notwaldorf:
(yeah this is required as per https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-356362524)

bicknellr:
I wonder why this is in Polymer's 3.x branch then?

PR #1107, opened 2018-04-28
[3.0-preview] Update Polymer CLI version

abdonrd:


abdonrd:
Friendly ping to @bicknellr. 😄

package.json
```diff
@@ -25,7 +25,7 @@
     "@webcomponents/webcomponentsjs": "^1.2.0"
   },
   "devDependencies": {
-    "polymer-cli": "^1.7.0-pre.13",
-    "wct-browser-legacy": "^0.0.1-pre.12"
+    "polymer-cli": "1.7.0-pre.15",
```
notwaldorf: why no caret?

abdonrd: Because is a `pre` version. Possible breaking changes.

Do you prefer add the caret?

notwaldorf: yeah, we've been caret-ing and also bumping the pre versions  ¯\_(ツ)_/¯

abdonrd: Done!

bicknellr:
There seems to be an issue with the distributed version of the CLI that's causing both this and #1103 to fail linting. Building the CLI package individually out of the tools repo also has this issue but building all of the tools repo together doesn't. I suspect this has something to do with the `npm-shrinkwrap.json` in that package, but I'm not sure yet.

abdonrd:
Yay! Happy green Travis!

PR #1109, opened 2018-05-08
[3.0-preview] Update polymer.json for webcomponentsjs changes.

bicknellr:
The polyfill bundles have moved into `/bundles`.

### DIFF
--- a/polymer.json
+++ b/polymer.json
@@ -7,7 +7,8 @@
   "extraDependencies": [
     "manifest.json",
     "node_modules/@webcomponents/webcomponentsjs/*.js",
-    "!node_modules/@webcomponents/webcomponentsjs/gulpfile.js"
+    "!node_modules/@webcomponents/webcomponentsjs/gulpfile.js",
+    "node_modules/@webcomponents/webcomponentsjs/bundles/*.js"
   ],
   "builds": [
     {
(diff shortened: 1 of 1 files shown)


bicknellr:
Actually, no, this needs to be updated to make sure that the builds include the new location of the webcomponentsjs bundles!

abdonrd:
I think with these we also fix it:
https://github.com/webcomponents/webcomponentsjs/pull/929#issuecomment-387305470

bicknellr:
@abdonrd is that because `polymer build` includes things listed in `files` in your build output?

abdonrd:
@bicknellr no, the idea is reduce the `npm install @webcomponents/webcomponentsjs` output.

You can see the actual output here:
https://unpkg.com/@webcomponents/webcomponentsjs@2.0.0-beta.2/

With https://github.com/webcomponents/webcomponentsjs/pull/929 we reduce the output to only:

```
"bundles/**/*",
"entrypoints/**/*",
"custom-elements-es5-adapter.js",
"webcomponents-bundle.js",
"webcomponents-bundle.js.map",
"webcomponents-loader.js"
```

bicknellr:
Oh, I read "I think with these we also fix it:" and though you were referring to this bug, rather than Keanu's comment about copying more than necessary.

abdonrd:
@bicknellr ops! Sorry for the confusion! 😅

PR #1112, opened 2018-05-08
[3.0-preview] Rename the global object used to store `rootPath` because it no longer affects Polymer.

bicknellr:


bicknellr:
(Note to self: the original implementation in 3da9aab10bf2e2f82f9ad5c3e89df175e14bd1c8 was preferable but broken when built. Still needs investigation.)

PR #1116, opened 2018-05-09
Updates for Polymer 3. (v4.0.0 prereleases)

bicknellr:
(I'll tag and release v4.0.0 after this is in.)


package.json
```diff
@@ -1,15 +1,34 @@
...
+    "@polymer/iron-selector": "^3.0.0-pre.15",
+    "@polymer/paper-icon-button": "^3.0.0-pre.15",
+    "@polymer/polymer": "^3.0.0",
+    "@webcomponents/webcomponentsjs": "^2.0.0"
+  },
+  "devDependencies": {
+    "polymer-cli": "^1.7.0",
+    "wct-browser-legacy": "^1.0.0"
+  },
+  "engines": {
```
keanulee: Do we need `"engines"`?

src/my-icons.js
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import '@polymer/iron-iconset-svg/iron-iconset-svg.js';
+const $_documentContainer = document.createElement('div');
```
keanulee: These should be `<template>`s now (and `.appendChild($_documentContainer.content);` below) due to template polyfill limitations on IE. See https://github.com/Polymer/shop/blob/3.0/src/shop-form-styles.js

src/my-app.js
```diff
@@ -0,0 +1,177 @@
...
+        type: String,
+        reflectToAttribute: true,
+        observer: '_pageChanged'
+      },
+      routeData: Object,
+      subroute: Object
+    };
+  }
+
+  ready() {
```
abdonrd: We can remove it.

bicknellr:
I *think* we're going to go ahead with 4.0.0 even though the elements are still in prerelease, but I'm double checking with some other folks right now to make sure.

package.json
bicknellr: I think this was added because the CLI doesn't support Node.js <8 any more.

src/my-icons.js
bicknellr: Thanks for pointing this out; fixed!

package.json
```diff
@@ -1,15 +1,34 @@
...
+    "@polymer/iron-iconset-svg": "^3.0.0-pre.15",
+    "@polymer/iron-media-query": "^3.0.0-pre.15",
+    "@polymer/iron-pages": "^3.0.0-pre.15",
+    "@polymer/iron-selector": "^3.0.0-pre.15",
+    "@polymer/paper-icon-button": "^3.0.0-pre.15",
+    "@polymer/polymer": "^3.0.0",
+    "@webcomponents/webcomponentsjs": "^2.0.0"
+  },
+  "devDependencies": {
+    "polymer-cli": "^1.7.0",
```
keanulee: Should use 1.7.1 which significantly reduces deps required.

PR #1117, opened 2018-05-10
Update package name to @polymer/starter-kit

RobertJGabriel:
Saw this issue and thought it made sense :)

https://github.com/Polymer/polymer-starter-kit/issues/1111


RobertJGabriel:
Just tagging @bicknellr  as I know this can get missed with emails and such

PR #1118, opened 2018-05-13
Navigation use enhancements.

legowerewolf:
These modifications make it easier to insert and remove pages from the app, because pages only have to be listed in a single place. It manages adding elements to the iron-pages element and the navigation menu.
* Pages with a non-null, non-false display field show up in the navigation menu.
* The page with "home" in the role field is loaded by default if there is no path.
* The page with the "err404" role is loaded when the path doesn't match an existing page.

### DIFF
--- a/src/my-app.js
+++ b/src/my-app.js
@@ -38,27 +38,27 @@ class MyApp extends PolymerElement {
         :host {
           --app-primary-color: #4285f4;
           --app-secondary-color: black;
-
+      
           display: block;
         }
-
+      
         app-drawer-layout:not([narrow]) [drawer-toggle] {
           display: none;
         }
-
+      
         app-header {
           color: #fff;
           background-color: var(--app-primary-color);
         }
-
+      
         app-header paper-icon-button {
           --paper-icon-button-ink-color: white;
         }
-
+      
         .drawer-list {
           margin: 0 20px;
         }
-
+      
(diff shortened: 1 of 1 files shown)


legowerewolf:
I signed it!

src/my-app.js
```diff
@@ -153,20 +178,7 @@ class MyApp extends PolymerElement {
...
-        import('./my-view2.js');
-        break;
-      case 'view3':
-        import('./my-view3.js');
-        break;
-      case 'view404':
-        import('./my-view404.js');
-        break;
-    }
+    import(this.pages[this.pages.map(a => a.name).indexOf(page)].file)
```
keanulee: This cannot be statically analyzed and `polymer build` won't work.

jsilvermist: But would this work ok if you just add it as a fragment in `polymer.json`? Is there anything else we would be relying on analysis for here?

legowerewolf: I can tell you that it `polymer serve`'s just fine, but I did notice that `polymer lint` complains that it can't analyze it. 
EDIT: This does seem like it may be related to [this issue](https://github.com/Polymer/polymer-linter/issues/96).
EDIT2: And [this one](https://github.com/Polymer/polymer-linter/issues/32).

potagit:
Signed it, FYI

PR #1132, opened 2019-03-28
Upgrade all dependencies to their latest version

yveslange:
Upgrading all the dependencies in package.json


yveslange:
I signed it!

keanulee:
IMO package.json specifies the "minimum requirements" for an app. You can still update dependencies by refreshing package-lock.json, but package.json shouldn't change unless your app actually needs a feature or bug fix from a specific version.

